PR #1, opened 2022-11-16
RSDK-875 migrate rdk proto methods

maximpertsov:
JIRA: https://viam.atlassian.net/browse/RSDK-875

Initial release of typescript SDK - packages all web proto methods used by RDK: https://github.com/viamrobotics/rdk/pull/1601

TODO:
* [ ] set test workflow branch to `@main` before merging


.eslintrc.js
```diff
@@ -0,0 +1,74 @@
+module.exports = {
```
edaniels, line 1: @micheal-parks we should embark on a way to single source these from a repo akin to https://github.com/pion/.goassets

.github/workflows/publish.yml
```diff
@@ -0,0 +1,27 @@
+name: NPM Publish
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  publish:
```
edaniels, line 9: this should rely on a new CI workflow that runs the same actions as PR

README.md
```diff
@@ -1 +1,11 @@
-# viam-typescript-sdk
\ No newline at end of file
+# Viam TypeScript SDK
+
+## (In)stability Notice
+
+> **Warning**
+> This is a beta release of the Viam TypeScript SDK. Stability is not guaranteed. Breaking changes are likely to occur, and occur often.
```
edaniels, line 6: let's not say beta here at first. this is definitely alpha at this point

buf.yaml
```diff
@@ -0,0 +1,10 @@
+version: v1
+breaking:
+  use:
+    - FILE
+lint:
+  use:
+    - DEFAULT
+deps:
```
edaniels, line 8: there should be a buf lock to commit too I believe

package.json
```diff
@@ -0,0 +1,52 @@
+{
+  "name": "@viamrobotics/sdk",
+  "version": "0.0.1",
+  "description": "",
+  "main": "./dist/main.umd.js",
+  "module": "./dist/main.es.js",
```
edaniels, line 6: did you verify this works well with rdk/npm start?

src/gen/common/v1/common_pb.d.ts
```diff
@@ -0,0 +1,474 @@
+// package: viam.common.v1
```
edaniels, line 1: can we avoid committing all generated files since the publish will do that work for us?

tsconfig.node.json
```diff
@@ -0,0 +1,11 @@
+{
```
edaniels, line 1: did you test this out in node?

vite.config.ts
```diff
@@ -0,0 +1,28 @@
+import { defineConfig } from "vite";
+import path from "node:path";
+import dts from "vite-plugin-dts";
+
+// https://vitejs.dev/config/
```
edaniels, line 5: what are we using this config for?

edaniels (changes requested):

package.json
maximpertsov: yes

README.md
maximpertsov:
```suggestion
> This is an alpha release of the Viam TypeScript SDK. Stability is not guaranteed. Breaking changes are likely to occur, and occur often.
```

vite.config.ts
maximpertsov: this is the config for vite to build the packages and type definitions

tsconfig.node.json
maximpertsov: i did not - this config might not be necessary for this package since we are focusing on web support. I added it here to be consistent with our other frontend repos. this is another thing we can centralize as you alluded to here: https://github.com/viamrobotics/viam-typescript-sdk/pull/1#discussion_r1026594211

vite.config.ts
edaniels: But where do we use vite here?

maximpertsov: ah, it's a little roundabout - vite is called by `npm run build` (https://github.com/viamrobotics/viam-typescript-sdk/blob/b1abc17a7a96bcfe1dbfd4e60d5220cef9de676d/package.json#L15), which is called by `make build`

src/gen/common/v1/common_pb.d.ts
maximpertsov: yeah - we should keep track of the files `main.ts` references directly (so just the `esm` files that rollup generates) and ignore the rest

vite.config.ts
micheal-parks: Vite will generate that nice UMD + ES module combo

vite.config.ts
```diff
@@ -0,0 +1,28 @@
...
+export default defineConfig({
+  define: {
+    "process.env.NODE_ENV": '"production"',
+  },
+  plugins: [dts()],
+  build: {
+    minify: true,
+    target: 'esnext',
+    lib: {
```
micheal-parks, line 14: ^^ Specifically here

.eslintrc.js
micheal-parks: Yeah we should probably publish our eslint / typescript rules [as an NPM package](https://eslint.org/docs/latest/developer-guide/shareable-configs) and add them as deps to the rest of our projects. Just made a ticket for myself.

Makefile
```diff
@@ -0,0 +1,19 @@
+PATH_WITH_TOOLS="`pwd`/node_modules/.bin:${PATH}"
+
+install:
+	npm ci --audit=false --legacy-peer-deps
```
micheal-parks, line 4: Why do we need `--legacy-peer-deps`? Was some dependency chain complaining?

maximpertsov: i think i added this by mistake - i don't see any issues when i remove this configuration. good catch

maximpertsov:
```suggestion
	npm ci --audit=false
```

micheal-parks: Nice! Yeah, unfortunately we have it in PRIME becuase storybook's dependency chain is a mess...

src/gen/common/v1/common_pb.d.ts
edaniels: we dont even need to commit that. no one clones this to pull those in. they can use the npm publish

maximpertsov:
sorry @micheal-parks @edaniels, I did some git history rewrite nonsense and closed the branch by accident - reopening here: https://github.com/viamrobotics/viam-typescript-sdk/pull/2

PR #2, opened 2022-11-18
RSDK-875 migrate rdk proto methods

maximpertsov:
JIRA: https://viam.atlassian.net/browse/RSDK-875

Initial release of typescript SDK - packages all web proto methods used by RDK: https://github.com/viamrobotics/rdk/pull/1601

TODO:
* [x] set test workflow branch to `@main` before merging


micheal-parks:
Could we add separate ESLint and TS jobs in CI like we do for the rest of our repos? A good example of this would be the `pullrequest.yml` workflow within the PRIME repo.

edaniels:
I'd like to stop doing that. It increases concurrency and adds unnecessary load. I removed it in rdk and those checks are fast anyway in an SDK like this so let's keep it the same

edaniels (approved):
LGTM

.eslintrc.js
```diff
@@ -0,0 +1,74 @@
+module.exports = {
```
maximpertsov, line 1: @micheal-parks I copied this as-is from three-inspect - let me know if any there are any settings that don't apply here.

micheal-parks (approved):

.github/workflows/pullrequest.yml
```diff
@@ -0,0 +1,11 @@
+name: Pull Request Update
+
+on:
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  test:
+    # TODO: lock to @main branch before merging
```
maximpertsov, line 10:
```suggestion
```

PR #4, opened 2022-11-23
Add a little more wrapping

edaniels:


maximpertsov (changes requested):
bump package to 0.0.3

looks good otherwise

maximpertsov (approved):

PR #5, opened 2022-11-30
RSDK-914 - Add session support

edaniels:
Depends on https://github.com/viamrobotics/api/pull/121
Depended on by https://github.com/viamrobotics/rdk/pull/1630

### DIFF
--- a/src/SessionManager.ts
+++ b/src/SessionManager.ts
@@ -89,6 +89,10 @@ export default class SessionManager {
       })
     }
 
+    /*
+     * This lint is correct but it makes our lives easier to refer to a boolean in
+     * case in the future we make this toggleable (e.g. foreground).
+     */
     // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
     if (this.backgroundHeartbeat && window.Worker) {
       const url = window.URL.createObjectURL(timeoutBlob)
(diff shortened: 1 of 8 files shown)


src/SessionManager.ts
```diff
@@ -0,0 +1,166 @@
...
+
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+    if (this.backgroundHeartbeat && window.Worker) {
+      const url = window.URL.createObjectURL(timeoutBlob)
+      worker = new Worker(url)
+      URL.revokeObjectURL(url)
+      worker.onmessage = function () {
+        doHeartbeat()
+      }
+    }
```
maximpertsov, line 100: is the lint exception a false positive?

edaniels: it is. it comes from the fact that it's not aware that you can have one value, do an await (yielding control to a different function), and come back and have that change. It assumes this functions runs all the way through

maximpertsov: got it - might be helpful to add that as a comment to explain the exception

maximpertsov (approved):

src/SessionManager.ts
edaniels: oh i lied. no it's because i have a bool always set to true to use background workers. if we set it to false in the future, it disables this code block here. just easier. ill explain in a comment

maximpertsov: thanks!

PR #6, opened 2022-12-08
RSDK-910 camera wrappers

maximpertsov:
JIRA: https://viam.atlassian.net/browse/RSDK-910

* Implement camera client and wrapper methods
* Add `promisify` utility to convert grpc callbacks to promises

Testing:
* [x] Replaced existing wrappers in RDK

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -58,6 +58,7 @@ module.exports = {
     'no-continue': 'off',
     'no-extra-parens': 'off',
     'no-magic-numbers': 'off',
+    'no-shadow': 'off', // https://github.com/typescript-eslint/typescript-eslint/issues/2483#issuecomment-687095358
     'no-ternary': 'off',
     'no-undefined': 'off',
     'object-curly-spacing': ['error', 'always'],
@@ -70,5 +71,6 @@ module.exports = {
     'semi': ['error', 'never'],
 
     '@typescript-eslint/prefer-for-of': 'off',
+    '@typescript-eslint/no-shadow': ['error'],
   },
 }
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -71,6 +71,6 @@ module.exports = {
     'semi': ['error', 'never'],
 
     '@typescript-eslint/prefer-for-of': 'off',
-    '@typescript-eslint/no-shadow': ['error'],
+    '@typescript-eslint/no-shadow': ['error'], // https://github.com/typescript-eslint/typescript-eslint/issues/2483#issuecomment-687095358
   },
 }
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@viamrobotics/sdk",
-  "version": "0.0.7",
+  "version": "0.0.8",
   "description": "",
   "main": "./dist/main.umd.js",
   "module": "./dist/main.es.js",
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
   "types": "./dist/main.d.ts",
   "files": [
     "dist/Client.*",
+    "dist/components/**/*.d.ts",
     "dist/main.*",
     "dist/gen/**/*.esm.d.ts",
     "dist/gen/**/*.d.ts"
--- a/src/Client.ts
+++ b/src/Client.ts
@@ -21,14 +21,14 @@ import { VisionServiceClient } from './gen/service/vision/v1/vision_pb_service.e
 import type { grpc } from '@improbable-eng/grpc-web'
 
 interface WebRTCOptions {
-  enabled: boolean;
-  host: string;
-  signalingAddress: string;
-  rtcConfig: RTCConfiguration | undefined;
+  enabled: boolean
+  host: string
+  signalingAddress: string
+  rtcConfig: RTCConfiguration | undefined
 }
 
 interface SessionOptions {
-  disabled: boolean;
+  disabled: boolean
 }
 
 export default class Client {
@@ -67,9 +67,7 @@ export default class Client {
 
   private movementSensorServiceClient: MovementSensorServiceClient | undefined
 
-  private inputControllerServiceClient:
-    | InputControllerServiceClient
-    | undefined
+  private inputControllerServiceClient: InputControllerServiceClient | undefined
 
   private motorServiceClient: MotorServiceClient | undefined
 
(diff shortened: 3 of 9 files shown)


src/Client.ts
```diff
@@ -230,6 +220,21 @@ export default class Client {
...
+    if (!clientTransportFactory) {
+      throw new Error(Client.notConnectedYetStr)
+    }
+    const grpcOptions = { transport: clientTransportFactory }
+    return {
+      grpcOptions,
+      serviceHost: this.serviceHost,
+    }
+  }
+
```
maximpertsov, line 237: this exposes connection parameters so that component service clients can instantiate themselves given a `Client`, rather than making the `Client` responsible for instantiating each component client itself.

src/components/camera/Client.ts
```diff
@@ -0,0 +1,79 @@
...
+export class CameraClient implements Camera {
+  private client: Client | undefined
+  private readonly name: string
+
+  constructor (client: Client, name: string) {
+    this.client = client
+    this.name = name
+  }
+
+  private get cameraService () {
```
Kschappacher, line 17: Im curious, what is the use case for this method?

package.json
```diff
@@ -1,12 +1,13 @@
 {
   "name": "@viamrobotics/sdk",
-  "version": "0.0.7",
+  "version": "0.0.8",
```
Kschappacher, line 3: will we have to update the version every push to main?

src/components/camera/Client.ts
maximpertsov: related to https://github.com/viamrobotics/viam-typescript-sdk/pull/6#discussion_r1044666570 - this method encapsulates creating a `CameraServiceClient` from an existing client connection.

package.json
maximpertsov: nope, only if we want to publish a new version. the underlying workflow we are using does not publish if we don't bump the version: https://github.com/JS-DevTools/npm-publish

src/components/camera/Client.ts
maximpertsov: I will add a utility for this logic since each component wrapper will need it - assuming we are okay sticking with this pattern

.eslintrc.js
```diff
@@ -70,5 +71,6 @@ module.exports = {
     'semi': ['error', 'never'],
 
     '@typescript-eslint/prefer-for-of': 'off',
+    '@typescript-eslint/no-shadow': ['error'], // https://github.com/typescript-eslint/typescript-eslint/issues/2483#issuecomment-687095358
```
micheal-parks, line 74: Ah, this is a good fix that we should propagate to the rest of our repos. Another reason to add a repo for our ESLint rules 😎

src/components/camera/Client.ts
```diff
@@ -0,0 +1,79 @@
+import { type Camera, MimeType, Properties } from './Camera'
+import { CameraServiceClient } from '../../gen/component/camera/v1/camera_pb_service.esm'
+import type Client from '../../Client'
+import type { HttpBody } from '../../gen/google/api/httpbody_pb'
+import pb from '../../gen/component/camera/v1/camera_pb.esm'
+import { promisify } from '../../utils'
+
+export class CameraClient implements Camera {
+  private client: Client | undefined
```
micheal-parks, line 9: Wanted to mention that JS has native private class members nowadays if we want to enforce these rules in both JS and TS: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields

src/components/camera/Client.ts
Kschappacher: I think it makes sense to either add it to utility or to have a component interface that has a generic version of this. Then we can include any shared component functions in the component interface and have new components get it for free. Something like this https://github.com/viamrobotics/viam-python-sdk/blob/main/src/viam/components/component_base.py

src/components/camera/Client.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  private get cameraService () {
+    if (!this.client) {
+      throw new Error('not connected yet')
+    }
+    const { grpcOptions, serviceHost } = this.client.serviceConnection
+    return new CameraServiceClient(serviceHost, grpcOptions)
+  }
+
+  async getImage (mimeType: MimeType): Promise<Uint8Array> {
```
Kschappacher, line 25: We should probably add tests for all of these methods using a mock camera. I think its fine that we dont have any in this pr since this is the first wrapper but they should be added in the future.

maximpertsov: For sure - I'm working on setting up unit tests and mocks in a separate branch: https://github.com/viamrobotics/viam-typescript-sdk/tree/RSDK-915-add-testing-to-ts-sdk-ci-pipeline.

Kschappacher (approved):
LGTM

src/main.ts
```diff
@@ -1,4 +1,11 @@
+import * as Camera from './components/camera/Camera'
+import { CameraClient } from './components/camera/Client'
 import Client from './Client'
+
+// eslint-disable-next-line no-warning-comments
+// TODO(RSDK-1017): remove this export once RDK stops using it directly
+import { RobotService } from './gen/robot/v1/robot_pb_service.esm'
```
edaniels, line 7: you can kill this here and now

src/Client.ts
```diff
@@ -354,7 +361,6 @@ export default class Client {
       this.armServiceClient = new ArmServiceClient(this.serviceHost, grpcOptions)
       this.baseServiceClient = new BaseServiceClient(this.serviceHost, grpcOptions)
       this.boardServiceClient = new BoardServiceClient(this.serviceHost, grpcOptions)
-      this.cameraServiceClient = new CameraServiceClient(this.serviceHost, grpcOptions)
```
edaniels, line 357: don't we still need a good way to ask the Client for a camera? Like `client.camera("name")`?

edaniels: Well we do that in go due to type limitations. In typescript you could technically pass enough type info to know you need to pull a camera clients. Also the wrappers are so light that I'm wondering if we prioritize the Dx here

maximpertsov: > In typescript you could technically pass enough type info to know you need to pull a camera clients.

I added [this generic service method](https://github.com/viamrobotics/viam-typescript-sdk/pull/6/commits/7d97641f68c30fb88dc9e26fbb12f026e4213b99) that is meant to be called by the component clients, but it could be called in the way you describing to cut down on some code. 

> Also the wrappers are so light that I'm wondering if we prioritize the Dx here

That's fair, the wrappers did turn out to be on the lighter side. I'm still leaning towards not having to import everything into client and making it so users only need to import what they need.

micheal-parks (approved):

PR #7, opened 2022-12-15
[RSDK-912 | RSDK-913] Motor and Base Wrappers

Kschappacher:
This pr includes the Base and Motor wrappers for the TS proto methods.


src/components/base/Client.ts
```diff
@@ -0,0 +1,91 @@
...
+import type { Base } from './Base'
+import { BaseServiceClient } from '../../gen/component/base/v1/base_pb_service.esm'
+import { promisify } from '../../utils'
+
+
+export class BaseClient implements Base {
+  private client: BaseServiceClient
+  private name: string
+
+  constructor (client: Client, name:string) {
```
maximpertsov, line 12: note to self / nit: our lint step should be catching these sort of minor issues (i.e. adding a space after `name:` in this case)

maximpertsov: hrm, I'm actually get a bunch of linting errors - curious why they aren't showing up in CI

src/components/motor/Motor.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export interface Motor {
+  setPower(power: number, extra?:Map<string, any>):Promise<motorApi.SetPowerResponse>;
+  goFor(rpm:number, revolutions:number, extra?:Map<string, any>):Promise<motorApi.GoForResponse>;
+  goTo(rpm:number, positionRevolutions:number, extra?:Map<string, any>):Promise<motorApi.GoToResponse>;
+  resetZeroPosition(offset:number, extra?:Map<string, any>):Promise<motorApi.ResetZeroPositionResponse>;
+  motorStop(extra?:Map<string, any>) :Promise<motorApi.StopResponse>;
+  getProperties(extra?:Map<string, any>) :Promise<Properties>;
+  getPosition(extra?:Map<string, any>):Promise<number>;
+  isPowered(extra?:Map<string, any>):Promise<(number | boolean)[]>;
```
maximpertsov, line 16:
```suggestion
  isPowered(extra?:Map<string, any>):Promise<[boolean, number]>;
```

maximpertsov (changes requested):
we should bump the version to `0.0.9`

src/components/base/Client.ts
maximpertsov: don't worry about this, [I'm making a separate branch to tighten up some of our linting rules](https://github.com/viamrobotics/viam-typescript-sdk/tree/update-lint-build). also, my errors were related to generating protobufs with another tool - I'll address that situation in the aforementioned branch as well.

src/components/motor/Motor.ts
Kschappacher: I tried doing Promise<[boolean, number]> but the type checker didnt like it. I think it has something to do with me returning an array of return values and the fact that technically the array could contain a number or a bool in either position.

Kschappacher:
> we should bump the version to `0.0.9`

Bumped!

src/components/motor/Client.ts
```diff
@@ -0,0 +1,123 @@
...
+  async isPowered (extra = {}) {
+    const motorService = this.motorService
+    const request = new motorApi.IsPoweredRequest()
+    request.setName(this.name)
+    request.setExtra(googleProtobufStructPb.Struct.fromJavaScript(extra))
+    const response = await promisify<motorApi.IsPoweredRequest, motorApi.IsPoweredResponse>(
+      motorService.isPowered.bind(motorService),
+      request
+    )
+    return [response.getIsOn(), response.getPowerPct()]
```
maximpertsov, line 120:
```suggestion
    return [response.getIsOn(), response.getPowerPct()] as const
```

src/components/motor/Motor.ts
Kschappacher: by golly you've done it

PR #9, opened 2022-12-28
Add stream client

maximpertsov:
Expose methods to add and remove a stream. We don't have a defined API streaming yet, but this feels like a good starting point for making it easy to setup streams.

JIRA: https://viam.atlassian.net/browse/RSDK-1019

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@viamrobotics/sdk",
-  "version": "0.0.9",
+  "version": "0.0.10",
   "description": "",
   "main": "./dist/main.umd.js",
   "module": "./dist/main.es.js",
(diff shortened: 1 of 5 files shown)


edaniels (approved):
LGTM as a temp thing since ideally this would be performed by the camera stream API

maximpertsov:
> LGTM as a temp thing since ideally this would be performed by the camera stream API

just to confirm, the camera proto does not have a stream api yet, right?

maximpertsov:
> > LGTM as a temp thing since ideally this would be performed by the camera stream API
> 
> just to confirm, the camera proto does not have a stream api yet, right?

Yeah looks like we don't - I agree it should totally be part of that once we add it.

edaniels:
It doesn't need a stream api proto to make a good client API. Camera API here in TS can call the stream methods for you. I don't think we will add those to the camera API anyway. That's for streaming frames, not starting a webrtc track

PR #13, opened 2023-01-17
Board sensor robot

Kschappacher:


src/components/robot/Client.ts
```diff
@@ -0,0 +1,175 @@
+import type Client from '../../Client'
```
maximpertsov, line 1: Let's remove the robot interface and client for now

Kschappacher: Done

maximpertsov (approved):

PR #14, opened 2023-01-18
RSDK-1740, RSDK-1745 add request logging / idiomatic responses

maximpertsov:
JIRA: https://viam.atlassian.net/browse/RSDK-1740, https://viam.atlassian.net/browse/RSDK-1745

* Adds client config option to log grpc requests
* Updates some wrapper return signatures to return primitive values instead of protobuf responses

### DIFF
--- a/src/components/base/Client.ts
+++ b/src/components/base/Client.ts
@@ -31,11 +31,10 @@ export class BaseClient implements Base {
 
     this.options.requestLogger?.(request)
 
-    const response = await promisify<baseApi.MoveStraightRequest, baseApi.MoveStraightResponse>(
+    await promisify<baseApi.MoveStraightRequest, baseApi.MoveStraightResponse>(
       baseService.moveStraight.bind(baseService),
       request
     )
-    return response
   }
 
   async spin (angleDeg: number, degsPerSec: number, extra = {}) {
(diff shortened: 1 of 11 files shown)


src/components/base/Client.ts
```diff
@@ -26,11 +29,12 @@ export class BaseClient implements Base {
     request.setDistanceMm(distanceMm)
     request.setExtra(googleProtobufStructPb.Struct.fromJavaScript(extra))
 
-    const response = await promisify<baseApi.MoveStraightRequest, baseApi.MoveStraightResponse>(
+    this.options.requestLogger?.(request)
```
Kschappacher, line 32: Im curious, what happens when requestLogger is not set and this line is called. will it break? or just not run?

maximpertsov: the `?.()` syntax makes it so we only execute the call if requestLogger is set

maximpertsov: but testing on controls page now to verify

Kschappacher: follow up question, has this been tested in local? and do we need any .bind to make this work?

Kschappacher (approved):
LGTM

src/components/base/Client.ts
maximpertsov: looks good in local, no need to bind here

PR #17, opened 2023-02-02
Use prettier to format

maximpertsov:


.eslintrc.js
```diff
@@ -9,8 +9,9 @@ module.exports = {
     'eslint:all',
     'plugin:@typescript-eslint/recommended',
     'plugin:@typescript-eslint/strict',
+    'prettier',
```
maximpertsov, line 12: this plugin turns off conflicting eslint rules, so we would still be able to import a shared eslint config.

micheal-parks (approved):

PR #20, opened 2023-02-04
RSDK-1843 Add wrapper for robot

maximpertsov:
JIRA: https://viam.atlassian.net/browse/RSDK-1843

This PR mainly adds a wrapper for the viam robot, but also introduces a few other changes:
* [x] `SessionManager` uses aforementioned robot wrapper
* [x] Documentation
* [x] Tests (only for Session Manager at the moment)

TODO:
* [x] Revert using `@main` branch for CI
* [x] Q on frame system: https://github.com/viamrobotics/viam-typescript-sdk/pull/20/files#r1100213286


src/components/robot/Client.ts
```diff
@@ -0,0 +1,234 @@
...
+};
+
+export class RobotClient implements Robot {
+  private client: RobotServiceClient;
+
+  constructor(init: RobotClientInit) {
+    if (isClient(init)) {
+      this.client = init.createServiceClient(RobotServiceClient);
+    } else if (isAddressInit(init)) {
+      this.client = new RobotServiceClient(init.serviceHost, init.opts);
```
maximpertsov, line 37: alternative constructor arguments for session manager context

.github/workflows/pullrequest.yml
```diff
@@ -7,4 +7,5 @@ on:
 
 jobs:
   test:
-    uses: viamrobotics/viam-typescript-sdk/.github/workflows/test.yml@main
+    # TODO: revert to main branch
```
maximpertsov, line 10: well revert once the PR is approved

src/components/robot/Client.ts
```diff
@@ -0,0 +1,234 @@
...
+    const response = await promisify<
+      proto.TransformPoseRequest,
+      proto.TransformPoseResponse
+    >(robotService.transformPose.bind(robotService), request);
+    const result = response.getPose();
+    if (!result) {
+      // eslint-disable-next-line no-warning-comments
+      // TODO: Can the response frame be undefined or null?
+      throw new Error('no pose');
+    }
```
maximpertsov, line 119: I'm not sure if getting a `null | undefined` pose is valid or should be an error.

.github/workflows/test.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - name: Build and lint
+      - name: Build, lint, test
```
edaniels, line 16:
```suggestion
      - name: Build, lint, and test
```

src/components/robot/Robot.ts
```diff
@@ -0,0 +1,195 @@
...
+  /**
+   * Get the list of operations currently running on the robot.
+   *
+   * @privateRemarks
+   *   TODO: this function should return an more idiomatic type instead of just
+   *   passing along a proto type.
+   * @group Operations
+   * @alpha
+   */
+  getOperations(): Promise<proto.Operation[]>;
```
edaniels, line 20: returning proto types isn't inherently a bad thing. The operation is a POD, so unless its API is really bad, this is O.K.

src/components/robot/Robot.ts
```diff
@@ -0,0 +1,195 @@
...
+   * @privateRemarks
+   *   TODO: this function should return an more idiomatic type instead of just
+   *   passing along a proto type.
+   * @param resume - A session ID that is used to attempt to continue a stream
+   *   after a disconnection event. If the session is not found, a new one will
+   *   be created and returned.
+   * @group Sessions
+   * @alpha
+   */
+  startSession(resume: string): Promise<proto.StartSessionResponse>;
```
edaniels, line 134: this session stuff bloats the interface. I'd move it to a session manager interface suitable for end users

src/components/robot/Client.ts
```diff
@@ -0,0 +1,234 @@
...
+    } else {
+      throw new Error('invalid arguments');
+    }
+  }
+
+  private get robotService() {
+    return this.client;
+  }
+
+  // OPERATIONS
```
edaniels, line 47: [nit] These types of comments aren't really useful

edaniels (changes requested):
Nice! Just a few comments. I'm assuming session manager did not change w.rt. fixes/functionality/flow other than adding the async versions of the functions (thanks for that!).

src/components/robot/Robot.ts
maximpertsov: what is "POD"?


i can see it both ways. my main concern is coupling the proto api to the sdk api, especially since we are interested in potentially moving to another protobuf generator at some point. but yeah, it's also annoying to define interfaces that are essentially pass-throughs.

not necessarily something we need to commit on yet, but should as we move to a stable release

src/components/robot/Client.ts
maximpertsov: I'd prefer to keep them for now if that's ok - I found these cheap comment sections helpful in the python SDK.

maybe this is a sign that robot is becoming a bit of god class / api?

maximpertsov:
> I'm assuming session manager did not change w.rt. fixes/functionality/flow other than adding the async versions of the functions

Correct, the tests I added passed against the original SessionManager implementation - I did not have have to update them at all after updating.

src/components/robot/Client.ts
maximpertsov: @DTCurrie let me know if this what you had in mind when you suggested making the constructor take a discriminated union.

maximpertsov: nevermind, we don't need this anymore

src/components/robot/Robot.ts
maximpertsov: agreed - it's gone now - only session manager exposes session stuff

src/components/robot/Client.ts
edaniels: I don't really think so. It's fairly small.

src/components/robot/Robot.ts
edaniels: POD is a plain ol datatype.

edaniels (approved):
LGTM

src/components/robot/Client.ts
maximpertsov: i asked a few people about this, and the loose consensus was that `null` pose should be an error

PR #21, opened 2023-02-13
Add stream client track events

micheal-parks:
This PR adds a generic event dispatcher class and hooks it up WebRTC track events. The `StreamClient` now allows users to subscribe to track events directly:

```ts
const client = new StreamClient()

client.on('track', event => {
  // do something with streams
})
```

This will help with organizing client state management since the SDK will no longer attempt to manage DOM element attributes.


src/extra/stream/Client.ts
```diff
@@ -1,17 +1,21 @@
...
+export class StreamClient extends EventDispatcher implements Stream {
   private client: StreamServiceClient;
   private readonly options: Options;
 
   constructor(client: Client, options: Options = {}) {
+    super();
     this.client = client.createServiceClient(StreamServiceClient);
     this.options = options;
+    
+    events.on('track', (args) => this.emit('track', args));
```
edaniels, line 18: can you add a test for this?

maximpertsov (approved):

src/extra/stream/Client.ts
micheal-parks: Yep! can do.

maximpertsov:
Feel free to remove the the `npm check` step from lint, it's complaining that our `rpc` library is not up to date - i'm going to make this a non-blocking CI step

PR #28, opened 2023-02-28
RSDK-2050 Update Arm Wrapper 

oliviamiller:
Added a TS Pose object, make worldstate parameter optional, changed movetoJointPostions parameter to a list, changed functions to lowercase.


src/components/arm/Client.ts
```diff
@@ -40,13 +40,32 @@ export class ArmClient implements Arm {
...
       throw new Error('no pose');
     }
-    return result;
+    return {
+      x: result.getX(),
+      y: result.getY(),
+      z: result.getZ(),
+      ox: result.getX(),
+      oy: result.getY(),
+      oz: result.getZ(),
```
maximpertsov, line 49:
```suggestion
      ox: result.getOX(),
      oy: result.getOY(),
      oz: result.getOZ(),
```

src/components/arm/Arm.ts
```diff
@@ -1,43 +1,45 @@
-import type { Pose, WorldState } from '../../gen/common/v1/common_pb.esm';
+import type { WorldState } from '../../gen/common/v1/common_pb.esm';
```
maximpertsov, line 1: ideally we should define our own type for world state like we did for pose, assuming it's not too complicated of a definition

src/components/arm/Arm.ts
```diff
@@ -1,43 +1,45 @@
...
-  MoveToPosition: (
+  moveToPosition: (
     pose: Pose,
-    world: WorldState,
+    world?: WorldState,
     extra?: Extra
   ) => Promise<void>;
 
   /**
    * Move each joint of the arm based on the angles on the joint poisitons
```
maximpertsov, line 25:
```suggestion
   * Move each joint of the arm based on the angles on the joint positions.
```

maximpertsov (changes requested):

maximpertsov:
We also need to bump to `0.0.23`

src/components/arm/Arm.ts
maximpertsov: discussed IRL, this struct is a bit more complicated so we can punt on this for now. there are a few other methods that expose protobuf structs in their signature so we can address when we change those

maximpertsov (changes requested):
lgtm! just need to bump it to version `0.0.25` and you're good

maximpertsov (approved):

PR #29, opened 2023-03-04
RSDK-1847 Return plain types instead of protobufs

maximpertsov:
JIRA: https://viam.atlassian.net/browse/RSDK-1847

* Updates return types of several wrappers to stop returning proto objects.
* Enforces prettier formatting.

Corresponding RDK wrapper update: https://github.com/viamrobotics/rdk/pull/1985


njooma (commented):
Is there a reason to redefine types as native? I'm of the opinion that we actually shouldn't do that, because now we have to keep our native definitions in sync with our proto definitions, and the whole point of doing proto is that it's language agnostic. 

It's something I notice in RDK as well, where we have to have so many utility functions just to go between native and proto. Seems redundant and introduces an additional point of failure.

But I guess that, as this is for a typed language, we'd hopefully catch any errors in CI (or earlier). But still, seems like unnecessary work to redefine types that already exist?

maximpertsov:
> Is there a reason to redefine types as native? I'm of the opinion that we actually shouldn't do that, because now we have to keep our native definitions in sync with our proto definitions, and the whole point of doing proto is that it's language agnostic.
> 
> It's something I notice in RDK as well, where we have to have so many utility functions just to go between native and proto. Seems redundant and introduces an additional point of failure.
> 
> But I guess that, as this is for a typed language, we'd hopefully catch any errors in CI (or earlier). But still, seems like unnecessary work to redefine types that already exist?

I think converting to native types gives us decoupling and ergonomics. The latter might not be as big of an improvement when we change to a different protobuf generator, but the current protobuf types are very confusing for end-users to work with.

njooma (approved):

PR #30, opened 2023-03-06
RSDK-1711 smarter dial logic

maximpertsov:
JIRA: https://viam.atlassian.net/browse/RSDK-1711

This PR makes two changes that brings the typescript SDK into closer parity with our other SDKs:

* Implement `createRobotClient`, which will be the preferred method for instantiating a robot client. This function attempts to dial the robot via WebRTC and then falls back to dialing via gRPC.
  * Update the teleop example to use this method as well.
* Merge `RobotClient` functionality into `Client`, effectively making the robot client.


examples/teleop/package-lock.json
```diff
@@ -10,7 +10,7 @@
       "license": "Apache-2.0",
       "dependencies": {
         "@improbable-eng/grpc-web": "^0.15.0",
-        "@viamrobotics/sdk": "^0.0.21"
+        "@viamrobotics/sdk": "file:../../../typescript-sdk/viamrobotics-sdk-0.0.27.tgz"
```
maximpertsov, line 13: Unfortunately, I will need to push a follow-up commit after this PR is merged and the new version is live.

examples/teleop/package-lock.json
```diff
@@ -82,9 +82,10 @@
       }
     },
     "node_modules/@viamrobotics/sdk": {
-      "version": "0.0.21",
-      "resolved": "https://registry.npmjs.org/@viamrobotics/sdk/-/sdk-0.0.21.tgz",
-      "integrity": "sha512-pyTcsYvc4aeupZbuvhnnLCEdjCtQvYVuSO6F6PGOpVIiBnu3X8kuSu8OeawCuUs7awgMqEVgiZf4IWuKA6KFeg==",
+      "version": "0.0.27",
+      "resolved": "file:../../../typescript-sdk/viamrobotics-sdk-0.0.27.tgz",
+      "integrity": "sha512-ZQek54dlYfj/3drz8wDN042V7WS7oVZpyqv4DgUa0/w2ZbEA0AoN0BCohx+E7nwA1PwH4yYZAsmRKbSW0CnMPg==",
```
maximpertsov, line 87: [ditto](https://github.com/viamrobotics/viam-typescript-sdk/pull/30/files#r1127011792)

edaniels:
@maximpertsov do you want me to review this?

maximpertsov:
> @maximpertsov do you want me to review this?

if you have the time then absolutely!

src/robot/dial.ts
```diff
@@ -0,0 +1,122 @@
+import { RobotClient } from './Client';
+
+interface DialDirectConf {
+  authEntity?: string;
+  host: string;
+  locationSecret?: string;
```
edaniels, line 6: how does one dial something that is not viam managed and with other credentials? locationSecret is one such type. Might it be better to more closely match the go options?

edaniels:
> > @maximpertsov do you want me to review this?
> 
> if you have the time then absolutely!

Time is but a construct.

src/robot/dial.ts
```diff
@@ -0,0 +1,122 @@
...
+}
+
+/** Check if a url corresponds to a local connection via heuristic */
+const isLocalConnection = (url: string) => url.includes('.local');
+
+const dialDirect = async (conf: DialDirectConf): Promise<RobotClient> => {
```
edaniels, line 12: I think we need to either export these or expose the right options so we can better connect.

src/robot/dial.ts
```diff
@@ -0,0 +1,122 @@
...
+  return client;
+};
+
+type Conf = DialDirectConf | DialWebRTCConf;
+
+const isDialWebRTCConf = (value: Conf): value is DialWebRTCConf => {
+  const conf = value as DialWebRTCConf;
+
+  if (typeof conf.signalingAddress !== 'string') return false;
+  if (!(conf.iceServers instanceof Array)) return false;
```
edaniels, line 92: ice servers are not strictly necessary to connect

src/robot/dial.ts
```diff
@@ -0,0 +1,122 @@
...
+
+export const createRobotClient = async (conf: Conf): Promise<RobotClient> => {
+  let client;
+
+  // Try to dial via WebRTC first.
+  if (isDialWebRTCConf(conf)) {
+    try {
+      client = await dialWebRTC(conf);
+    } catch (err) {
+      console.warn('failed to connect via WebRTC...');
```
edaniels, line 105: we may want to drop all of these into debug for now

edaniels (changes requested):
Left a few comments. Main thing I think we need is some config parity with go but more importantly to bring back credentials as a config option.

src/robot/dial.ts
maximpertsov: fair, i did this to make dialing viam more convenient but this is too deeply coupled at this point

edaniels: yeah and we want to make sure we always are supporting our OSS cases well in addition to other types of credentials that an app managed robot can already accept (like a JWT)

maximpertsov: i was considering letting this function accept a union of interfaces that could support these different use cases, but that maybe that's a bit confusing...

src/robot/dial.ts
maximpertsov: adjusted to reflect this (took an embarrassing number of commits...)

examples/teleop/src/index.js
```diff
@@ -122,11 +94,9 @@ connectWebRTC()
...
       console.log('linear', linear);
       console.log('angular', angular);
-      // TODO: stop requiring users to import commonApi, if possible
-      const linearVec = new commonApi.Vector3();
-      const angularVec = new commonApi.Vector3();
-      linearVec.setY(linear);
-      angularVec.setZ(angular);
+
+      const linearVec = { x: 0, y: linear, z: 0 };
+      const angularVec = { x: 0, y: 0, z: angular };
```
maximpertsov, line 99: not related to any change in this PR, but needed to adjust to fit an api change from an earlier version bump

src/robot/dial.ts
maximpertsov: exported the conf types: https://github.com/viamrobotics/viam-typescript-sdk/pull/30/commits/d2344febb12d134f9da95c4b65439898d787ed06

i think most editors will still give hints even if we don't export, but doesn't hurt

edaniels (approved):
LGTM

PR #32, opened 2023-03-22
RSDK-2433 App code sample

maximpertsov:
JIRA: https://viam.atlassian.net/browse/RSDK-2433

* Add code samples that to hydrate typescript's code samples tab
* [flyby-ish] Add `isMoving` methods for base and motor, since they are showcased
* [flyby] Export data manager client

Depends on https://github.com/viamrobotics/app/pull/1484 - merge that first


codesamples/apis.json
maximpertsov, line 1: We are including components/services for which we have wrappers.

src/services/data-manager/data-manager.ts
maximpertsov, line 1: We are going to change our linting rules to enforce kebab-casing for files, so this is in the preparation for that.

Will do the rest in a subsequent PR.

codesamples/apis.json
```diff
@@ -0,0 +1,46 @@
...
+    "movement_sensor": {
+        "client": "MovementSensorClient",
+        "func": "getLinearAcceleration",
+        "args": []
+    },
+    "motion": {
+        "client": "MotionClient",
+        "func": "getPose",
+        "args": [],
+        "comment": "Note this function requires additional arguments. Please provide valid arguments for function to work properly."
```
maximpertsov, line 39: we are doing this for python as well

Kschappacher (approved):
LGTM

PR #37, opened 2023-03-27
RSDK-2183 - Query for valid SDP names

cheukt:
tested in conjunction with https://github.com/viamrobotics/rdk/pull/2115
also works with RDKs on main.
Deploy order is approximately TS SDK -> RC -> APP -> actual change in RDK

### DIFF
--- a/src/robot/Client.ts
+++ b/src/robot/Client.ts
@@ -42,7 +42,7 @@ interface SessionOptions {
 }
 
 abstract class ServiceClient {
-  constructor(public serviceHost: string, public options?: grpc.RpcOptions) {}
+  constructor(public serviceHost: string, public options?: grpc.RpcOptions) { }
 }
 
 /** A gRPC-web client for the Robot component. */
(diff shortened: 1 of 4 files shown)


edaniels (approved):

src/robot/Client.ts
```diff
@@ -325,14 +325,26 @@ export class RobotClient implements Robot {
...
+          // where names should not contain colons.
+          const resName = eventStream.id.replaceAll('+', ':');
+          // overriding the stream id to match the resource name and then immediately setting it back to readonly
+          Object.defineProperty(eventStream, 'id', {
+            value: resName,
+            writable: true,
+          });
+          Object.defineProperty(eventStream, 'id', {
+            writable: false,
+          });
```
maximpertsov, line 348: nice find! just curious, can we change the property and make it read-only in one step? or does `read-only` prevent this kind of direct hacking?

```suggestion
            writable: false,
          });
```

src/robot/Client.ts
```diff
@@ -325,14 +325,26 @@ export class RobotClient implements Robot {
...
 
           const eventStream = event.streams[0];
           if (!eventStream) {
+            events.emit('track', event);
             throw new Error('expected event stream to exist');
           }
+
+          // track id has +s to conform to RFC 4566 (https://www.rfc-editor.org/rfc/rfc4566)
+          // where names should not contain colons.
+          const resName = eventStream.id.replaceAll('+', ':');
```
maximpertsov, line 340: nit: let's put `getValidSDPTrackName` into utils and call it here too just so the logic is centralized?

maximpertsov (changes requested):
i like this approach - just some minor change requests + can we add a test for this?

src/robot/Client.ts
cheukt: huh I was having trouble getting to work before but apparently it doesn't even need the `writable: false`

src/robot/Client.ts
cheukt: this does the reverse and I feel like these should actually be private functions - we don't want users to need to interact with these functions

maximpertsov (commented):
one more thing, we need to bump to `0.0.30` - i released a new version just a second ago

maximpertsov (approved):

PR #38, opened 2023-03-28
Remove legacy media dom injection logic

maximpertsov:


cheukt (approved):
LGTM if tested against a current version of the RC

maximpertsov:
> LGTM if tested against a current version of the RC

yes will do

PR #41, opened 2023-03-31
Reduce build size by 92%

micheal-parks:
Reduces build size from ~25mb to ~2mb and removes the need to run rollup during `make buf`. ~Need to test with RDK.~ Max and I have tested by locally linking to RDK, and Max has tested on a real robot. We haven't detected any issues.


edaniels:
Can you describe what the main idea was to bring the size down? Was it optimize deps? Trying to sort through the various changes to understand

micheal-parks:
> Was it optimize deps

@edaniels Yep! It's that and `commonjsOptions`. Not entirely sure if it will works perfectly though. Will test Monday morning by linking it to the RDK.

edaniels:
Sweet

src/components/arm/Client.ts
```diff
@@ -1,11 +1,11 @@
-import * as googleProtobufStructPb from 'google-protobuf/google/protobuf/struct_pb';
+import { Struct } from 'google-protobuf/google/protobuf/struct_pb';
```
maximpertsov, line 1: style change or affects bundle size?

src/components/arm/Client.ts
```diff
@@ -1,11 +1,11 @@
...
 import type { RobotClient } from '../../robot';
-import pb from '../../gen/component/arm/v1/arm_pb.esm';
-import { ArmServiceClient } from '../../gen/component/arm/v1/arm_pb_service.esm';
+import pb from '../../gen/component/arm/v1/arm_pb';
```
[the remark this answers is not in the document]
maximpertsov: ah that's right... got it

maximpertsov (approved):

src/robot/SessionManager.test.ts
```diff
@@ -26,7 +26,7 @@ const mockHealthyHeartbeat = (_req, _md, cb) => {
 describe('SessionManager', () => {
   beforeEach(() => {
     sm = new SessionManager(host, transport);
-    vi.mock('./gen/robot/v1/robot_pb_service.esm');
+    vi.mock('./gen/robot/v1/robot_pb_service');
```
mcous, line 29: Not a change introduced by this PR, but `vi.mock` gets hoisted to the top of the file and runs only once. It should probably be moved out of the `beforeEach` to make that clear

(You could also replace it with `vi.doMock`, which will not be hoisted and run on every `beforeEach`, but due to vitest's module caching, this won't do anything other than potentially slow the test suite down)

mcous (commented):
I don't feel like I have the context in our codebase yet to officially approve this, but it LGTM!

I was confused by the extra rollup config in this repo, so dropping that and configuring CommonJS > ESM conversion through vite directly seems like a much better solution

PR #44, opened 2023-04-03
Cleanup test mocks

maximpertsov:
Update: now uses conventions established in other tests - see https://github.com/viamrobotics/viam-typescript-sdk/pull/141 for related discussion.

---

Clean up mocks for existing test

Context: https://github.com/viamrobotics/viam-typescript-sdk/pull/41/files#r1156191232

### DIFF
--- a/src/robot/session-manager.test.ts
+++ b/src/robot/session-manager.test.ts
@@ -1,10 +1,11 @@
 // @vitest-environment happy-dom
 
-import { afterEach, beforeEach, describe, expect, test, vi } from 'vitest';
+import { beforeEach, describe, expect, test, vi } from 'vitest';
 import { ConnectionClosedError } from '@viamrobotics/rpc';
 import { FakeTransportBuilder } from '@improbable-eng/grpc-web-fake-transport';
 import { grpc } from '@improbable-eng/grpc-web';
 import { RobotServiceClient } from '../gen/robot/v1/robot_pb_service';
+vi.mock('../gen/robot/v1/robot_pb_service');
 
 import SessionManager from './session-manager';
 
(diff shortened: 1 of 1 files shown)


src/robot/SessionManager.test.ts
```diff
@@ -26,7 +26,7 @@ const mockHealthyHeartbeat = (_req, _md, cb) => {
 describe('SessionManager', () => {
   beforeEach(() => {
     sm = new SessionManager(host, transport);
-    vi.mock('./gen/robot/v1/robot_pb_service');
+    vi.doMock('./gen/robot/v1/robot_pb_service');
```
mcous, line 29: This path doesn't match the import above.

```ts
import { RobotServiceClient } from '../gen/robot/v1/robot_pb_service';
// ...
vi.doMock('./gen/robot/v1/robot_pb_service');
```

Is this mock doing anything? Commenting the mock out does not change the test outcome in any way that I can see

maximpertsov: Indeed - we stub out the relevant methods accessed in each test, so there's no need to mock the whole module.

mcous (approved):
I figure this PR is fine to merge, since, the `vi.mock` call wasn't doing anything. That being said, it's my opinion that testing via mocking out individual methods of a class is substantially worse than mocking out the whole module.

Mocking out individual methods creates a [partial mock][] in the test. The docs I linked to do a good job of laying out why partial mocks are an anti-pattern that should be avoided.

I've been working on a package called [vitest-when][] that I'd like to pull in here at some point to provide a more focused, safer mocking API. I'm working through some TypeScript issues with overloaded functions at the moment, but I'll put a PR up once those have been resolved

[partial mock]: https://github.com/testdouble/contributing-tests/wiki/Partial-Mock
[vitest-when]: https://github.com/mcous/vitest-when

PR #45, opened 2023-04-05
Improve makefile and add contributing guide

mcous:
## Overview

While working on the React teleoperation example, I noticed a few problems with the Makefile:

- No documentation on how to use it
- Targets are not spec'd as `.PHONY`
- Variables use backticks for shell expansion rather than evaluating with `$(shell ...)`
- Targets were named inconsistently according to both internal and common Makefile conventions

This PR is an effort to clean some of that up to make external contribution easier

## Changelog

- Add basic contributing guide
- Mark all phony Makefile targets as `.PHONY`
- Fix `PATH` handling
- Explicitly use local `buf` executable rather than relying on it being present in `PATH`
- Rename `install` targets to `setup` to match `viamrobotics/rdk`
    - In Makefiles, `install` means "install the built artifact on the system so I can use it"
    - Here, we mean "install development dependencies so that I can develop"
    - Also added corresponding `teardown` targets to undo `setup`, which can be helpful
- Ensure `setup` targets only run if needed
- Add default `all` target to clean, build, lint, and test
- Rework targets to that they're parallelizable via `make -j`


CONTRIBUTING.md
```diff
@@ -0,0 +1,51 @@
...
+### Prerequisites
+
+You'll need to make sure the following tools are installed on your machine and accessible in your `PATH`.
+
+- Node.js
+  - `node`
+  - `npm`
+- Protobuf
+  - `protobuf`
+  - `protoc-gen-grpc-web`
```
mcous, line 22: I don't think this is extensive. I had to run the setup instructions from `viamrobotics/app` before I could get my local dev environment working, so I don't know what the minimum required setup is here

Makefile
```diff
@@ -1,51 +1,92 @@
...
+	rm -rf src/gen
 
-build: clean-dist buf
-	# TODO(RSDK-870): try removing the custom `--max-old-space-size` option
-	# once we migrate to protobuf-es, since that generator should produce
-	# much smaller javascript bundles.
-	NODE_OPTIONS="--max-old-space-size=16384" npm run build
+.PHONY: update-buf
+update-buf: $(buf)
+	$(buf) mod update
```
mcous, line 56: Does this target need to be a dependency of any other target? What does it do?

CONTRIBUTING.md
```diff
@@ -0,0 +1,51 @@
+# Contributing Guide
+
+Welcome to the Viam TypeScript SDK. Please read this guide prior to opening a pull request.
+
+## Development setup
+
+Developing on the TypeScript SDK requires macOS, Linux, or WSL / Bash for Windows. To get started, clone this repository:
```
mcous, line 7: Makefile uses `rm -rf`, which will wreck havoc on a Windows machine if you try to use `cmd.exe` or PowerShell

.github/workflows/update_protos.yml
```diff
@@ -3,7 +3,7 @@ name: Update Protos + Open PR
 on:
   push:
     branches:
-      - "RSDK-2091-auto-PR-api-changes"
+      - 'RSDK-2091-auto-PR-api-changes'
```
maximpertsov, line 6: we can delete this entire `push: ...` section actually - it was left over from testing

Makefile
maximpertsov: it updates the version of [our api](https://github.com/viamrobotics/api) that we generate protos from, so basically an api upgrade. it should not be a dependency of any targets.

Makefile
```diff
@@ -1,51 +1,92 @@
...
+.PHONY: clean-js
+clean-js:
+	rm -rf dist
+
+# TODO(RSDK-870): try removing the custom `--max-old-space-size` option
+# once we migrate to protobuf-es, since that generator should produce
+# much smaller javascript bundles.
+.PHONY: build-js
+build-js: $(node_modules) clean-js build-buf
+	NODE_OPTIONS="--max-old-space-size=16384" npm run build
```
maximpertsov, line 86:
```suggestion
.PHONY: build-js
build-js: $(node_modules) clean-js build-buf
	npm run build
```

another thing we can get rid of, thanks to https://github.com/viamrobotics/viam-typescript-sdk/pull/41

CONTRIBUTING.md
maximpertsov: hm, I'll see if we can stop relying on that. We might not need to remove `dist` since it since it gets cleared by vite. we do need to clear out generated protobufs tho.

CONTRIBUTING.md
maximpertsov: hm, curious what that was. the only other dependency I can think of is `buf`, but we install that one locally to `bin/buf`

maximpertsov: oh, looks like `buf` is a available through npm now - might be easier to use that

.github/workflows/update_protos.yml
maximpertsov: boo, GH isn't letting me select all 3 lines to delete as a suggestion :(

CONTRIBUTING.md
mcous: I think getting rid of `rm -rf` is a good goal, but probably not the highest priority. You're right that it isn't really needed for `vite`. I've used `rimraf` in the past, which works but introduces an annoying catch 22 where you can't clean without having your dev environment set up. Maybe that's ok though, especially if we can install `buf` via `npm`

mcous: I gave Windows support a little timebox, but the issues seem a little larger than just `rm` support. Punting on it for now, will file a ticket to address at a later date

src/components/arm/Arm.ts
maximpertsov, line 1: weird that this change is showing up in the diff

maximpertsov: oh nevermind i see we have a conflict, it's that

Makefile
```diff
@@ -1,24 +1,29 @@
-buf := ./bin/buf
 node_modules := $(shell npm root)
+buf := npm exec --offline -- buf
```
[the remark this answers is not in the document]
mcous: Interesting. This command is supposed to mean...

> Run `buf` from `node_modules` and error out if it's not available

The little `$(node_modules)` dependency may be a little too clever. An alternative would be to remove the dependency and rely on users to run `make setup` manually when needed, rather than try to do it automatically based on if `node_modules` is present `package-lock.json` hasn't changed

src/components/arm/Arm.ts
mcous: Uh I'm also confused why this is showing up in the diff. I'm gonna rebase because it's making me nervous, apologies in advance for the force push!

CONTRIBUTING.md
maximpertsov: agreed, we can tackle this later

Makefile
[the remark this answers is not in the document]
maximpertsov: i think we still need the dependency to ensure that these commands run in CI, since both require `buf` to be installed through npm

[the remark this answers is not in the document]
maximpertsov: I'm okay with the current setup - if it causes problems for local dev we can revisit.

maximpertsov (approved):

PR #51, opened 2023-04-06
RSDK-2037, RSDK-2571 Clean-up docs and add naked grpc examples

maximpertsov:
This PR makes a few documentation improvements:

* Set documentation homepage to be the code search index instead of the README.
* Mark raw grpc methods as deprecated in cases where we have wrappers.
* Add minimal examples of using raw/naked protobuf apis for components / services where we do not have wrappers:
image

JIRA
* https://viam.atlassian.net/browse/RSDK-2571
* https://viam.atlassian.net/browse/RSDK-2037


tsconfig.json
```diff
@@ -41,7 +41,7 @@
   "include": [
     "src/**/*.ts",
     "src/**/*.d.ts",
-    "src/**/*.vue"
+    "*.cjs"
```
maximpertsov, line 44: unrelated change - I changed our config files to be common js to appease the linter, and also removed the config for vue files since this project does not have any.

src/main.ts
```diff
@@ -1,72 +1,417 @@
...
+ * const request = new sensorsApi.GetSensorsRequest();
+ * request.setName('my sensors');
+ *
+ * client.sensorsService.getSensors(
+ *   request,
+ *   new grpc.Metadata(),
+ *   (error, response) => {
+ *     // do something with error or response
+ *   }
+ * );
```
maximpertsov, line 133: I'm only including this kind of example for components/services that do not have a wrapper yet - we should deprecate and remove the example once the corresponding wrapper is available.

src/main.ts
```diff
@@ -1,72 +1,417 @@
...
+ *
+ * @alpha
+ * @group Raw Protobufs
+ */
+export { default as gantryApi } from './gen/component/gantry/v1/gantry_pb';
+
+/**
+ * Raw Protobuf interfaces for a Generic component.
+ *
+ * Generated with https://github.com/improbable-eng/grpc-web
```
maximpertsov, line 184: I'm going to leave this one without an example until someone asks for it.

src/robot/Client.ts
```diff
@@ -130,6 +134,13 @@ export class RobotClient implements Robot {
     return this.robotServiceClient;
   }
 
+  /**
+   * Raw Protobuf service generated by
+   * https://github.com/improbable-eng/grpc-web for an Arm component.
+   *
+   * @deprecated Use {@link ArmClient} instead.
+   * @group Raw Protobufs
+   */
```
maximpertsov, line 143:
```suggestion
```

src/components/arm/Client.ts
```diff
@@ -7,7 +7,11 @@ import { promisify } from '../../utils';
 import type { Arm } from './Arm';
 import commonPB from '../../gen/common/v1/common_pb';
 
-/** A gRPC-web client for the Arm component. */
+/**
+ * A gRPC-web client for the Arm component.
+ *
+ * @group Clients
```
cheukt, line 13: what does @group mean here?

src/main.ts
cheukt: I think Generic is a good one to have an example for actually

src/main.ts
```diff
@@ -1,72 +1,417 @@
...
+ *
+ * @example
+ *
+ * ```ts
+ * import { grpc } from '@improbable-eng/grpc-web';
+ *
+ * const client = {}; // replace with a connected robot client
+ *
+ * const request = new sensorsApi.GetSensorsRequest();
+ * request.setName('my sensors');
```
cheukt, line 125: nit: these aren't valid names

cheukt (changes requested):

src/components/arm/Client.ts
maximpertsov: it groups them together in the online documentation

src/main.ts
maximpertsov: oops, that's more than a nit - will update

cheukt (approved):
LGTM

PR #52, opened 2023-04-06
Integrate common JS configs

mcous:
## Overview

This PR integrates the common ESLint, TypeScript, and Prettier configs from [viamrobotics/js-configs](https://github.com/viamrobotics/js-config) into the SDK project. Since this project is a vanilla TS library, it proved the easiest place to get started, compared to Prime (hybrid Vue + Svelte situation) or App (Vue + major refactors in progress).

All linting problems (183, at the time of writing) were downgraded to warnings and suppressed via `--quiet` to avoid overwhelming CI and introducing a bunch of churn via `eslint --fix`. I would address them in followup PR(s).

## Change log

- Update linting, formating, and typechecking config to use common configs
- Switch TypeScript build system from `vite-plugin-dts` to `tsc`
- Update package files list

### Detailed Build Changes

I ran into one issue with this project's build, though:

- This project was recently updated to use TypeScript v5
- TypeScript v5 deprecated the `isolatedModules` config in favor of `verbatimModuleSyntax`
- Our base config uses `verbatimModuleSyntax`
- This plugin uses `vite-plugin-dts` to generate .d.ts files for publish
- `vite-plugin-dts` is locked to TypeScript v4 by a dependency

I typically just use vanilla `tsc` to generate type definitions for libraries I publish, so I took a stab at implementing that setup here, replacing `vite-plugin-dts` with `tsc`. It's output looks equivalent from a file list and file size standpoint, as reported by `npm pack`.

Changes required to `tsconfig.json`:

- To take source files in `src` and spit out `.d.ts` files in `dist`, `tsconfig.rootDir` needs to be set to `src`
- If `tsconfig.rootDir` is set to `src`, TS complains if you import anything outside of `src`
- The library imports `package.json`, which is outside of `src`

To solve this last problem, I implemented something I've done in other packages: use Vite to compile the version string into the source via a obvious named constant - in this case: `__VERSION__`.

### Detailed file list changes

I changed the `package.json::files` field to `dist`. This caused the following 4 files to be added to the pack, where they weren't before:

```
npm notice 431B   dist/services/datamanager/client.d.ts
npm notice 119B   dist/services/datamanager/datamanager.d.ts
npm notice 424B   dist/types.d.ts
npm notice 752B   dist/utils.d.ts
```

From comparing the results of `npm pack`, all other files between this branch and `main` are identical.


maximpertsov (approved):
LGTM - happy to have a fewer tooling dependencies. It was probably unnecessary to include `vite-plugin-dts` and `tslib` in the first place.

PR #55, opened 2023-04-07
RSDK-1979 add motion service

maximpertsov:
JIRA: https://viam.atlassian.net/browse/RSDK-1979

* Add motion service wrapper.
* Shape wrapper interface to accept plain objects rather than protobuf objects - these should match the interface we get from protobufs' `AsObject` interface. Sadly these is no built-in `fromObject` method so we have to write functions for this.


src/utils.ts
```diff
@@ -40,10 +40,122 @@ export const encodeVector3D = function (value: Vector3D): pb.Vector3 {
...
 /** Convert a 3D Vector Protobuf Datatype to a POJO */
-export const decodeVector3D = function (proto: pb.Vector3): Vector3D {
+export const decodeVector3D = function (proto: common.Vector3): Vector3D {
   return {
     x: proto.getX(),
     y: proto.getY(),
     z: proto.getZ(),
   };
 };
+
```
maximpertsov, line 50: Most of the code below will go away once we migrate to protobuf-es since that library generate protobuf structures that are essentially plain objects. for now we need to have custom functions to convert from objects -> protos.

src/services/motion/types.ts
```diff
@@ -0,0 +1,54 @@
+import pb from '../../gen/service/motion/v1/motion_pb';
+
+export type MotionConstraints = pb.Constraints.AsObject;
+
```
maximpertsov, line 4: Most of the code below will go away once we migrate to protobuf-es since that library generate protobuf structures that are essentially plain objects. for now we need to have custom functions to convert from objects -> protos.

maximpertsov:
I tested these changes against a live arm. Unfortunately we actually use the motionApi yet so we might have to wait for it to be developed more to test that wrapper.

cheukt (approved):
LGTM

PR #56, opened 2023-04-11
Export typedefs for protos

maximpertsov:
We need to copy `*.d.ts` files that we generate over to dist, since `tsc` and `vite-plugin-dts` ignore any `.d.ts` files in source apparently.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,7 @@ teardown:
 
 .PHONY: build
 build: build-buf build-js
+	npm run copy-dts
 
 .PHONY: clean
 clean: clean-js clean-buf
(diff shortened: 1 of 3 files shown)


maximpertsov:
Confirming that proto typedefs are available with this package:

image

micheal-parks (approved):

package.json
```diff
@@ -24,6 +24,7 @@
   "scripts": {
     "build": "vite build",
```
mcous, line 25: Should this be part of the existing `build` script?

```suggestion
    "build": "vite build && npm run copy-dts",
```

Makefile
```diff
@@ -15,6 +15,7 @@ teardown:
 
 .PHONY: build
 build: build-buf build-js
+	npm run copy-dts
```
mcous, line 18: I would add this to the `build-js` target (or add it to the npm `build` script, as in my other comment)

mcous (approved):
LGTM, though I might move some stuff around slightly

Makefile
maximpertsov, line 18:
```suggestion
```

PR #58, opened 2023-04-12
RSDK-2480: Encoder API TypeScript SDK changes

martha-johnston:
This change adds a wrapper to the TypeScript SDK to support the new encoder component API.

### DIFF
--- a/src/components/encoder.ts
+++ b/src/components/encoder.ts
@@ -1,0 +1,2 @@
+export type { Encoder } from './encoder/Encoder';
+export { EncoderClient } from './encoder/Client';
--- a/src/components/encoder.ts
+++ b/src/components/encoder.ts
@@ -1,2 +1,2 @@
-export type { Encoder } from './encoder/Encoder';
+export type { Encoder, EncoderProperties } from './encoder/Encoder';
 export { EncoderClient } from './encoder/Client';
--- a/src/components/encoder/Client.ts
+++ b/src/components/encoder/Client.ts
@@ -52,21 +52,14 @@ export class EncoderClient implements Encoder {
       encoderApi.GetPropertiesRequest,
       encoderApi.GetPropertiesResponse
     >(encoderService.getProperties.bind(encoderService), request);
-    return {
-      ticksCountSupported: response.getTicksCountSupported(),
-      angleDegreesSupported: response.getAngleDegreesSupported(),
-    };
+    return response.toObject();
   }
 
-  async getPosition(positionType?: PositionType, extra = {}) {
+  async getPosition(positionType: PositionType = PositionType.POSITION_TYPE_UNSPECIFIED, extra = {}) {
     const encoderService = this.encoderService;
     const request = new encoderApi.GetPositionRequest();
     request.setName(this.name);
-    if (positionType === undefined) {
-      request.setPositionType(PositionType.POSITION_TYPE_UNSPECIFIED);
-    } else {
-      request.setPositionType(positionType);
-    }
+    request.setPositionType(positionType);
     request.setExtra(Struct.fromJavaScript(extra));
 
     this.options.requestLogger?.(request);
--- a/src/components/encoder/Client.ts
+++ b/src/components/encoder/Client.ts
@@ -1,0 +1,80 @@
+import { Struct } from 'google-protobuf/google/protobuf/struct_pb';
+import type { RobotClient } from '../../robot';
+import type { Encoder } from './Encoder';
+import { EncoderServiceClient } from '../../gen/component/encoder/v1/encoder_pb_service';
+import { type Options, PositionType } from '../../types';
+import encoderApi from '../../gen/component/encoder/v1/encoder_pb';
+import { promisify } from '../../utils';
+
+/**
+ * A gRPC-web client for the Encoder component.
+ *
+ * @group Clients
+ */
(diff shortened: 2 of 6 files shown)


randhid (commented):
Looks good on my end, will wai for the TSar.

src/components/encoder/Client.ts
```diff
@@ -0,0 +1,80 @@
...
+    this.options.requestLogger?.(request);
+
+    const response = await promisify<
+      encoderApi.GetPropertiesRequest,
+      encoderApi.GetPropertiesResponse
+    >(encoderService.getProperties.bind(encoderService), request);
+    return {
+      ticksCountSupported: response.getTicksCountSupported(),
+      angleDegreesSupported: response.getAngleDegreesSupported(),
+    };
```
maximpertsov, line 58: these protobuf structs have a built-in `.toObject()` method that will do this for you
```suggestion
    return response.toObject();
```

src/components/encoder/Encoder.ts
```diff
@@ -0,0 +1,25 @@
+import type { Extra, PositionType } from '../../types';
+
+export interface Properties {
+  /** Whether a encoder supports ticks and/or degrees. */
+  ticksCountSupported: boolean;
+  angleDegreesSupported: boolean;
+}
```
maximpertsov, line 7: we're gonna use a new pattern here (sorry for leading you astray with Motor!) - we can just re-export the `.AsObject` type from the underlying proto

```suggestion
import pb from '../../gen/service/encoder/v1/encoder_pb';

export const EncoderProperties = pb.Properties.AsObject;
```

See for an example of this: https://github.com/viamrobotics/viam-typescript-sdk/pull/55/files#diff-1aa6d58266a33208d7c59494dad01b893ebe3bc15fe369667d7554cc929e0297R3

(dealers choice on whether you create a dedicated `./types` folder or just keep the definition here - we can figure out structuring later).

src/components/encoder/Encoder.ts
```diff
@@ -0,0 +1,25 @@
...
+
+/** Represents a physical encoder. */
+export interface Encoder {
+  /** Set the current position of the encoder as the new zero position. */
+  resetPosition(extra?: Extra): Promise<void>;
+
+  /** Return the encoder's properties. */
+  getProperties(extra?: Extra): Promise<Properties>;
```
maximpertsov, line 15: if you agree with re-exporting the type

```suggestion
  getProperties(extra?: Extra): Promise<EncoderProperties>;
```

src/components/encoder.ts
```diff
@@ -0,0 +1,2 @@
+export type { Encoder } from './encoder/Encoder';
```
maximpertsov, line 1: if you agree with re-exporting the type

```suggestion
export type { Encoder, EncoderProperties } from './encoder/Encoder';
```

src/components/encoder/Encoder.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  /**
+   * Return the current position either in relative units (ticks away from a
+   * zero position) or absolute units (degrees along a circle).
+   */
```
maximpertsov, line 20: let's add a @param comment for the optional position type argument.

src/main.ts
```diff
@@ -71,6 +71,18 @@ export { type Board, BoardClient } from './components/board';
...
+ * Raw Protobuf interfaces for an Encoder component.
+ *
+ * Generated with https://github.com/improbable-eng/grpc-web
+ *
+ * @deprecated Use {@link EncoderClient} instead.
+ * @alpha
+ * @group Raw Protobufs
+ */
+export { default as encoderApi } from './gen/component/encoder/v1/encoder_pb';
+export { type Encoder, EncoderClient } from './components/encoder';
```
maximpertsov, line 84: if you agree with re-exporting the type

```suggestion
export { type Encoder, type EncoderProperties, EncoderClient } from './components/encoder';
```

src/components/encoder/Client.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  async getPosition(positionType?: PositionType, extra = {}) {
+    const encoderService = this.encoderService;
+    const request = new encoderApi.GetPositionRequest();
+    request.setName(this.name);
+    if (positionType === undefined) {
+      request.setPositionType(PositionType.POSITION_TYPE_UNSPECIFIED);
+    } else {
+      request.setPositionType(positionType);
+    }
```
maximpertsov, line 69: alternatively we can set `POSITION_TYPE_UNSPECIFIED` as the default value:

```suggestion
  async getPosition(positionType: PositionType = PositionType.POSITION_TYPE_UNSPECIFIED, extra = {}) {
    const encoderService = this.encoderService;
    const request = new encoderApi.GetPositionRequest();
    request.setName(this.name);
    request.setPositionType(positionType);
```

maximpertsov (changes requested):
nice! just a few suggestions (most are related to a type change suggestion)

src/components/encoder/Client.ts
maximpertsov: it's actually equivalent to what you have - it will use the default if you omit the value [e.g. `getPosition()`] OR if pass in `undefined` as the first arg [e.g. `getPosition(undefined)`].

maximpertsov (approved):

PR #62, opened 2023-04-13
Fix most lints

mcous:
## Overview

This PR picks off all the "minor" linting warnings after merging #52. That means anything that was purely stylistic or fixable via `eslint --fix`.

There are about 50 remaining warnings, mostly from `unicorn/filename-case` and various `@typescript-eslint` rules that are going to be more involved to fix.

## Test plan

Due to number of files touched, IMO this PR requires smoke testing with actual hardware. The riskiest move here is _probably_ the import reordering, since it could cause changes in behavior if we have any problematic circular dependencies.

### DIFF
--- a/src/robot/SessionTransport.ts
+++ b/src/robot/SessionTransport.ts
@@ -58,14 +58,15 @@ export default class SessionTransport implements grpc.Transport {
     this.sessionManager
       .getSessionMetadata()
       .then((md) => {
+        // eslint-disable-next-line unicorn/no-array-for-each
         md.forEach((key: string, values: string | string[]) => {
           metadata.set(key, values);
         });
         this.transport.start(metadata);
         this.mdPromResolve?.();
       })
-      .catch((err) => {
-        this.opts.onEnd(err);
+      .catch((error) => {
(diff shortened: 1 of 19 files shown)


src/robot/SessionTransport.ts
```diff
@@ -58,14 +58,15 @@ export default class SessionTransport implements grpc.Transport {
...
         md.forEach((key: string, values: string | string[]) => {
           metadata.set(key, values);
         });
         this.transport.start(metadata);
         this.mdPromResolve?.();
       })
-      .catch((err) => {
-        this.opts.onEnd(err);
+      .catch((error) => {
+        this.opts.onEnd(error);
```
maximpertsov, line 69: interesting, so is this `err -> error` rules specifically for the catch clause? i see `err` vars elsewhere in this PR that we unchanged. 

not a big deal and certainly not blocking, but feels a little inconsistent...

maximpertsov (approved):
The changes themselves seem safe, but I agree we should test against the control panel of a live bot.

mcous:
Superseded by #70

PR #64, opened 2023-04-20
Remove enum

maximpertsov:


src/components/encoder/Encoder.ts
```diff
@@ -1,8 +1,12 @@
-import type { Extra, PositionType } from '../../types';
+import type { Extra } from '../../types';
 import pb from '../../gen/component/encoder/v1/encoder_pb';
 
 export type EncoderProperties = pb.GetPropertiesResponse.AsObject;
 
+type ValueOf<T> = T[keyof T];
+export const EncoderPositionType = pb.PositionType;
+export type EncoderPositionType = ValueOf<typeof pb.PositionType>;
```
maximpertsov, line 8: I'm exposing a type and value with the same name `EncoderPositionType` that vary slightly. My intention is to make the type/value pair function like the original enum without being an enum. Please let me know if this is confusing or if there is a better way.

PR #70, opened 2023-04-27
Fix most lints++

maximpertsov:
Builds on https://github.com/viamrobotics/viam-typescript-sdk/pull/62 - the only new thing this adds is consistent filenames


maximpertsov:
As discussed in the initial linting PR, I did a smoke test with these changes against: https://app.viam.com/robot?id=5082f509-02a7-4933-a035-dc8ef0681bed (a basic viam rover) - all components worked as before and there were no JS errors.

PR #71, opened 2023-04-27
RSDK-2835 - Add DoCommand to all resource clients

cheukt:
tested on a fake arm in RC - thoughts on how to add tests without just mocking everything?

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -211,7 +211,7 @@ export { default as gantryApi } from './gen/component/gantry/v1/gantry_pb';
  *
  * const client = {}; // replace with a connected robot client
  *
- * const request = new genericApi.DoCommandRequest();
+ * const request = new commonApi.DoCommandRequest();
  * request.setName('mygeneric');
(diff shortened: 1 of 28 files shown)


maximpertsov:
> how to add tests without just mocking everything?

right now we would just have to mock unfortunately - in the case of the do command we aren't doing much logic aside from type conversions so your call on whether you want to add a test - I'm okay with excluding them.

medium-term i'm considering importing another SDK that implements server-side components and testing against that.

maximpertsov (approved):

src/main.ts
```diff
@@ -211,7 +211,7 @@ export { default as gantryApi } from './gen/component/gantry/v1/gantry_pb';
  *
  * const client = {}; // replace with a connected robot client
  *
- * const request = new genericApi.DoCommandRequest();
+ * const request = new commonApi.DoCommandRequest();
```
maximpertsov, line 214: unrelated to your changes - i just noticed this mistake while reviewing

PR #73, opened 2023-04-28
RSDK-2899 Build library as part of example installation

maximpertsov:
Our examples all reference typescript sdk locally - this changes ensure that gets built as part of the installation process.

### DIFF
--- a/examples/teleop-elm/package.json
+++ b/examples/teleop-elm/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "example",
+  "name": "@viamrobotics/elm-teleop-sdk-example",
   "description": "An example tele-operation application using @viamrobotics/sdk",
   "scripts": {
     "preinstall": "make -C ../.. build",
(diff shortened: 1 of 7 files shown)


examples/teleop-elm/package.json
```diff
@@ -1,8 +1,8 @@
 {
-  "name": "@viamrobotics/sdk-teleop-demo",
-  "version": "0.0.1",
-  "description": "Viam SDK tele-operation demo in Elm",
+  "name": "example",
+  "description": "An example tele-operation application using @viamrobotics/sdk",
   "scripts": {
+    "preinstall": "cd ../.. && make build",
```
mcous, line 5: I like using the `-C` option for this rather than manually `cd`'ing

```suggestion
    "preinstall": "make -C ../.. build",
```

mcous, line 5: Is there any reason to do this as `prestart`, instead? The build process isn't very long, and for local dev, this might be more ergonomic

mcous (approved):
LGTM. These example projects have been giving me some grief, and this seems like an improvement. One thought for future improvements (currently floating around as a TODO somewhere) is to investigate `npm` workspaces so we only need one, top-level `npm install`

examples/teleop-elm/package.json
maximpertsov: the build process isn't that long but still noticeable IMO - I'd prefer to do it less

PR #74, opened 2023-05-01
Fix arm#isMoving call

maximpertsov:


maximpertsov:
already fixed in last release: https://github.com/viamrobotics/viam-typescript-sdk/commit/cfb7df223be594db24c0c44c02e4a91b88de68b8 - thanks @cheukt !

PR #80, opened 2023-05-08
RSDK-3015 - Fix streamStatus

cheukt:
status stream used to not work at all, reworked so it works? maybe.

pretty open to suggestions as to where I should be putting the new class, and its name, but all it is is a concrete impl of ResponseStream leveraging the existing EventDispatcher class.

tested using the below block
```
  let stream = robot.streamStatus([], new Duration().setNanos(500_000_000));
  stream.on('data', (response: VIAM.robotApi.Status[]) => {
    console.log(response);
  });
  stream.on('status', (newStatus?: { details: unknown }) => {
    console.error('error streaming robot status', newStatus);
  });
  stream.on('end', () => {
    console.error('done streaming robot status');
  });
```

### DIFF
--- a/src/robot.ts
+++ b/src/robot.ts
@@ -1,4 +1,4 @@
-export type { Robot } from './robot/robot';
+export type { Robot, RobotStatusStream } from './robot/robot';
 export { RobotClient } from './robot/client';
 export {
   type DialConf,
--- a/src/robot/client.ts
+++ b/src/robot/client.ts
@@ -23,11 +23,13 @@ import { MotorServiceClient } from '../gen/component/motor/v1/motor_pb_service';
 import { MovementSensorServiceClient } from '../gen/component/movementsensor/v1/movementsensor_pb_service';
 import { NavigationServiceClient } from '../gen/service/navigation/v1/navigation_pb_service';
 import { RobotServiceClient } from '../gen/robot/v1/robot_pb_service';
+import type { ResponseStream, Status } from '../gen/robot/v1/robot_pb_service';
 import { SLAMServiceClient } from '../gen/service/slam/v1/slam_pb_service';
 import { SensorsServiceClient } from '../gen/service/sensors/v1/sensors_pb_service';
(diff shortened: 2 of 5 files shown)


src/robot/client.ts
```diff
@@ -589,15 +591,29 @@ export class RobotClient implements Robot {
...
     request.setResourceNamesList(resourceNames);
     request.setEvery(duration);
-    const response = await promisify<
-      proto.StreamStatusRequest,
-      proto.StreamStatusResponse
-    >(robotService.streamStatus.bind(robotService), request);
-    return response.getStatusList();
+
+    const statusStream = robotService.streamStatus(request);
+    if (statusStream === null) {
```
maximpertsov, line 604:
```suggestion
    if (!statusStream) {
```

i'm assuming we want this to behave similar to the logic we have in `promisify` right?

maximpertsov (changes requested):
This seems okay - just some minor changes.

In addition, let's codify your testing example as a unit test:

```ts
let stream = robot.streamStatus([], new Duration().setNanos(500_000_000));
  stream.on('data', (response: VIAM.robotApi.Status[]) => {
    console.log(response);
  });
  stream.on('status', (newStatus?: { details: unknown }) => {
    console.error('error streaming robot status', newStatus);
  });
  stream.on('end', () => {
    console.error('done streaming robot status');
  });
```

src/robot/client.ts
cheukt: ya

cheukt:
@maximpertsov this was tested using the sample code from app and adding the bit I wrote above

src/robot.ts
```diff
@@ -1,4 +1,4 @@
-export type { Robot } from './robot/robot';
+export type { Robot, RobotStatusStream } from './robot/robot';
```
maximpertsov, line 1: we should export this in `src/main.ts` as well so users have access to this type

PR #87, opened 2023-05-16
RSDK-3131 - Auto reconnect + stream restart

maximpertsov:
Reconnect and repopulate streams for WebRTC connections.

### Testing

* [Run teleop-elm example app](./examples/teleop-elm/README.md)
* Open app in browser and confirm video feed works
* Disable wifi and confirm that the video feed is frozen
* Re-enable wifi and confirm a new working video feed has replaced the old frozen one after a few seconds

### DIFF
--- a/examples/teleop-elm/src/index.js
+++ b/examples/teleop-elm/src/index.js
@@ -42,8 +42,7 @@ function onTrack(event) {
       mediaElement.controls = true;
     }
 
-    const child = streamContainer.querySelector('video');
-    child?.remove();
+    streamContainer.querySelector('video')?.remove();
     streamContainer.append(mediaElement);
   }
 }
--- a/examples/teleop-elm/src/index.js
+++ b/examples/teleop-elm/src/index.js
@@ -41,6 +41,9 @@ function onTrack(event) {
     } else {
       mediaElement.controls = true;
     }
+
+    const child = streamContainer.querySelector('video');
+    child?.remove();
     streamContainer.append(mediaElement);
   }
 }
--- a/examples/teleop-elm/src/index.js
+++ b/examples/teleop-elm/src/index.js
@@ -14,7 +14,6 @@ async function connectWebRTC() {
     },
     authEntity: host,
     signalingAddress,
-    // noReconnect: true,
   });
 }
 
--- a/examples/teleop-elm/src/index.js
+++ b/examples/teleop-elm/src/index.js
@@ -14,6 +14,7 @@ async function connectWebRTC() {
     },
     authEntity: host,
     signalingAddress,
+    // noReconnect: true,
   });
 }
 
--- a/src/extra/stream/client.ts
+++ b/src/extra/stream/client.ts
@@ -22,13 +22,13 @@ const getValidSDPTrackName = (name: string) => {
 export class StreamClient extends EventDispatcher implements Stream {
   private client: StreamServiceClient;
   private readonly options: Options;
-  private streams: Record<string, boolean>;
+  private streams: Set<string>;
 
   constructor(client: RobotClient, options: Options = {}) {
     super();
     this.client = client.createServiceClient(StreamServiceClient);
     this.options = options;
-    this.streams = {};
+    this.streams = new Set();
 
     /**
      * Currently this is emitting events for every track that we recieve. In the
@@ -40,11 +40,8 @@ export class StreamClient extends EventDispatcher implements Stream {
     });
 
     events.on('reconnected', () => {
-      // TODO: add option to opt-out of stream restart?
-      for (const [name, active] of Object.entries(this.streams)) {
-        if (active) {
-          void this.add(name);
-        }
+      for (const name of this.streams.values()) {
+        void this.add(name);
       }
     });
   }
@@ -64,7 +61,7 @@ export class StreamClient extends EventDispatcher implements Stream {
         streamService.addStream.bind(streamService),
         request
       );
-      this.streams[name] = true;
+      this.streams.add(name);
     } catch {
       // Try again with just the resource name
       request.setName(name);
@@ -73,7 +70,7 @@ export class StreamClient extends EventDispatcher implements Stream {
         streamService.addStream.bind(streamService),
         request
       );
-      this.streams[name] = true;
+      this.streams.add(name);
     }
   }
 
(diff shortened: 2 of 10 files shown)


examples/teleop-elm/src/index.js
```diff
@@ -41,6 +41,9 @@ function onTrack(event) {
     } else {
       mediaElement.controls = true;
     }
+
+    const child = streamContainer.querySelector('video');
+    child?.remove();
```
maximpertsov, line 46: this makes it so new video feeds replace old ones, instead of being stacked below

src/robot/client.ts
```diff
@@ -336,6 +341,37 @@ export class RobotClient implements Robot {
...
          */
         this.peerConn = webRTCConn.peerConnection; // eslint-disable-line require-atomic-updates
+        this.peerConn.addEventListener('iceconnectionstatechange', () => {
+          /*
+           * TODO: are there any disconnection scenarios where we can reuse the
+           * same connection and restart ice?
+           *
+           * All connection loss scenarios I tested seem to result in the peer
+           * connection getting closed, so restarting ice is not a valid way to
+           * recover.
```
maximpertsov, line 351: @edaniels maybe you have some insight here - it seems like all connection loss scenarios will result in a peer connection close. I'm guessing this semantic is required for webrtc over grpc-web? 

Ideally I would like to reuse the existing peer connection, but obviously this is not possible once it's closed.

edaniels: I think our usage of ICE restart is very low and we almost always end up reconnecting. I think for now, we should reset our peer connection and make a new one when we've detected a disconnection. It's likely the server side need works to support ICE restart in addition to utilizing the established negotiation channel for SDP retransmission.

examples/teleop-elm/src/index.js
edaniels: can you shorten to `streamContainer.querySelector('video')?.remove()`?

src/robot/client.ts
```diff
@@ -336,6 +341,36 @@ export class RobotClient implements Robot {
...
+           */
+          if (this.peerConn?.iceConnectionState === 'closed') {
```
edaniels, line 353: you may want to only do this if we knew we 'connected' before

src/robot/client.ts
```diff
@@ -336,6 +341,36 @@ export class RobotClient implements Robot {
...
+                    `failed to reconnect - retries count: ${retries}`
+                  );
+                  retries += 1;
+                  throw error;
+                }
+              )
+            );
+          }
+        });
+
```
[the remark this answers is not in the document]
edaniels: I think there's a way you can have npm pack/build a prod version that removes debugs anyway such that you can leave helpful things around for future debugging sessions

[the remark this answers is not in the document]
edaniels: https://github.com/vitejs/vite/discussions/7920

src/robot/client.ts
```diff
@@ -336,6 +341,36 @@ export class RobotClient implements Robot {
...
+           *
+           * All connection loss scenarios I tested seem to result in the peer
+           * connection getting closed, so restarting ice is not a valid way to
+           * recover.
+           */
+          if (this.peerConn?.iceConnectionState === 'closed') {
+            let retries = 0;
+            console.debug('connection closed, will try to reconnect');
+            void backOff(() =>
+              this.connect().then(
```
edaniels, line 357: is connect ensuring no others can try connecting at the same time?

src/extra/stream/client.ts
```diff
@@ -22,18 +22,31 @@ const getValidSDPTrackName = (name: string) => {
...
+      this.emit('track', args);
+    });
+
+    events.on('reconnected', () => {
+      // TODO: add option to opt-out of stream restart?
+      for (const [name, active] of Object.entries(this.streams)) {
+        if (active) {
+          void this.add(name);
+        }
+      }
```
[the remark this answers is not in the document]
maximpertsov: will punt on making this configurable

src/robot/client.ts
maximpertsov: it's not, let me add that guard

maximpertsov: actually wait, I think we're okay since `connect` itself guards and exists early if we are connecting.

src/robot/client.ts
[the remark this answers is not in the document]
maximpertsov: that configuration doesn't work unfortunately. but I think it's fine to leave the logs in for now. will do make better configuration for removing these as a follow-up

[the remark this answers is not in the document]
edaniels: sounds good

src/extra/stream/client.ts
```diff
@@ -73,6 +88,7 @@ export class StreamClient extends EventDispatcher implements Stream {
         streamService.removeStream.bind(streamService),
         request
       );
+      this.streams[name] = false;
```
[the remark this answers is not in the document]
maximpertsov: make this a set instead https://github.com/viamrobotics/viam-typescript-sdk/pull/87/commits/1fbe2be5ebe4e25f5b8d25d6b3da74d587e70761

src/robot/client.ts
maximpertsov: unfortunately we can't get the previous state from the event. we could add a variable to track this, but maybe it's enough to rely on the fact that we set this listener right after we get the peer connection. we only reconnect on `close` so we know that peer connection will not accepts more requests after that.

maximpertsov:
Added an opt-out flag in case this feature causes issues: https://github.com/viamrobotics/viam-typescript-sdk/pull/87/commits/b026738ece8ec064ef3857154c8200df4c04b5a5

edaniels (approved):
LGTM

examples/teleop-elm/src/index.js
benjirewis: > instead of being stacked below

Can you explain what this means and why it is not a good thing?

maximpertsov: > Can you explain what this means and why it is not a good thing?

Discussed IRL - the old code would result in a stale video element on top of a working video element - we want to remove the stale video element before inserting the new working one.

PR #92, opened 2023-05-23
wip: use vitest-when to stub rather than partially mock

mcous:
Work in progress, see https://github.com/viamrobotics/viam-typescript-sdk/pull/44#pullrequestreview-1435020821


mcous:
Closing to keep things clean. Might come back to this in the future

PR #99, opened 2023-05-26
Add encoder api to codesamples

randhid:
Hopefully, this is all I'll need to do here. 😆

### DIFF
--- a/codesamples/apis.json
+++ b/codesamples/apis.json
@@ -26,6 +26,11 @@
         "func": "getImage",
         "args": []
     },
+    "encoder": {
+        "client": "EncoderClient",
+        "func": "getProperties",
+        "args": []
+    },
     "sensor": {
         "client": "SensorClient",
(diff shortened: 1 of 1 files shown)


martha-johnston:
Does this also add encoders to the Golang tab? If I recall that was missing encoders as well

randhid:
> Does this also add encoders to the Golang tab? If I recall that was missing encoders as well

No -that's in the rdk PR here: examples/apis.json

martha-johnston (approved):

maximpertsov (approved):
Yeah this will add the sample for typescript - if you want to test this out you can open a PR in app with [this line](https://github.com/viamrobotics/app/blob/3ff1ca2afdaa17e68b9e858356541c113897bc00/frontend/scripts/postinstall.js#L32) changed to point to this branch and check the change in a staging environment.

PR #119, opened 2023-06-12
RSDK-2866: add disconnection and reconnection events

purplenicole730:


src/robot/robot.ts
```diff
@@ -146,4 +148,15 @@ export interface Robot {
...
   ): RobotStatusStream;
+
+  /**
+   * Call a function when an event of either 'reconnected' or 'disconnected' is 
+   * triggered.
+   * 
+   * @param event - The event that was triggered.
+   * @param listener - The function to call
+   * @group Connection
+   * @alpha
```
purplenicole730, line 159: What `@group` should this be in, if any? What does `@alpha` do?

src/robot/client.ts
```diff
@@ -353,32 +389,7 @@ export class RobotClient implements Robot {
...
-                  console.debug(
-                    `failed to reconnect - retries count: ${retries}`
-                  );
-                  retries += 1;
-                  throw error;
-                }
-              )
-            );
-          }
+          this.isConnected() ? events.emit('reconnected', {}) : events.emit('disconnected', {});
```
purplenicole730, line 392: Lint error caused by this line:
`392:11  error  Expected an assignment or function call and instead saw an expression`

src/robot/robot.ts
maximpertsov: the group is purely for organization -  we can change this later or even leave it off. `Event Listeners` might be a good name if we end up adding non-connection events later.

src/robot/robot.ts
```diff
@@ -146,4 +148,15 @@ export interface Robot {
     resourceNames?: ResourceName.AsObject[],
     durationMs?: number
   ): RobotStatusStream;
+
+  /**
+   * Call a function when an event of either 'reconnected' or 'disconnected' is 
+   * triggered.
+   * 
+   * @param event - The event that was triggered.
```
maximpertsov, line 156: [nit] we might want to note the event types here.

Separately, we might also want to call out that these events only apply to WebRTC connections.

maximpertsov (changes requested):
looks great! some minor suggestions

src/robot/robot.ts
```diff
@@ -146,4 +149,19 @@ export interface Robot {
     resourceNames?: ResourceName.AsObject[],
     durationMs?: number
   ): RobotStatusStream;
+
+  /**
+   * Call a function when an event of either 'reconnected' or 'disconnected' is
+   * triggered. NOTE: Please note that these functions currently only apply to
+   * WebRTC connections.
```
maximpertsov, line 156:
```suggestion
   * triggered. Note that these events will only be triggered on WebRTC connections.
```

maximpertsov (approved):

PR #121, opened 2023-06-16
Rsdk 3333 add reconnect configurability

purplenicole730:


src/robot/client.ts
```diff
@@ -39,7 +39,10 @@ interface WebRTCOptions {
   host: string;
   signalingAddress: string;
   rtcConfig: RTCConfiguration | undefined;
+  // reconnection options
   noReconnect?: boolean;
+  reconnectMaxAttempts?: number;
+  reconnectMaxWait?: number;
```
cheukt, line 45: we should document what the default is and what valid inputs are

src/robot/client.ts
```diff
@@ -140,20 +143,32 @@ export class RobotClient extends EventDispatcher implements Robot {
...
+                );
+              }
+              throw error;
+            }
+          ),
+        {
+          // default values taken from `exponential-backoff` library
+          maxDelay:
+            this.webrtcOptions?.reconnectMaxWait || Number.POSITIVE_INFINITY,
+          numOfAttempts: this.webrtcOptions?.reconnectMaxAttempts || 10,
```
cheukt, line 170: is there an acceptable range for these numbers? we should probably check that maxDelay and numOfAttempts aren't negative, for example and throw an exception on client startup if configuration is wrong

cheukt (changes requested):

src/robot/client.ts
```diff
@@ -39,7 +39,12 @@ interface WebRTCOptions {
   host: string;
   signalingAddress: string;
   rtcConfig: RTCConfiguration | undefined;
+  // RECONNECTION OPTIONS
   noReconnect?: boolean;
+  // value should be a positive int; default is 10
```
cheukt, line 44: These comments should show up in docs (you can check using `make build` and looking in the docs folder), which I think requires the `/** */` syntax. we also only export `DialWebRTCConf` so the comments would be more suited to show up there

cheukt (changes requested):

cheukt (approved):
LGTM

PR #128, opened 2023-07-06
RSDK-3657: TS SDK Gantry API changes

martha-johnston:


src/components/gantry/client.ts
```diff
@@ -0,0 +1,144 @@
...
+    const response = await promisify<
+      pb.GetPositionRequest,
+      pb.GetPositionResponse
+    >(gantryService.getPosition.bind(gantryService), request);
+
+    const result = response.getPositionsMmList();
+    if (!result) {
+      throw new Error('no positions');
+    }
+    return result;
```
maximpertsov, line 46: linter says this will never be falsy so we can return without guarding:

```suggestion
    return response.getPositionsMmList();
```

src/components/gantry/client.ts
```diff
@@ -0,0 +1,144 @@
...
+      pb.GetLengthsRequest,
+      pb.GetLengthsResponse
+    >(gantryService.getLengths.bind(gantryService), request);
+
+    const result = response.getLengthsMmList();
+
+    if (!result) {
+      throw new Error('no lengths');
+    }
+    return result;
```
maximpertsov, line 109: linter says this will never be falsy so we can return without guarding:

```suggestion
    return response.getLengthsMmList();
```

src/components/gantry/client.ts
maximpertsov: @mcous - should we make `@typescript-eslint/no-unnecessary-condition` an error instead of a warning?

src/main.ts
```diff
@@ -198,7 +210,7 @@ export { type Stream, StreamClient } from './extra/stream';
  * @alpha
  * @group Raw Protobufs
  */
-export { default as gantryApi } from './gen/component/gantry/v1/gantry_pb';
+// export { default as gantryApi } from './gen/component/gantry/v1/gantry_pb';
```
maximpertsov, line 213: let's remove this and the doc-string above it (lines 185-213)

maximpertsov (changes requested):
there are some unnecessary checks and comments we can remove but otherwise lgtm

buf.lock
```diff
@@ -9,8 +9,8 @@ deps:
   - remote: buf.build
     owner: viamrobotics
     repository: api
-    commit: 77bcd1d0f47a437e8c58639004a13451
-    digest: shake256:9d311393e8b744ced31428355d42d3a997a92ee5cc9ecfa10cb30e656ad265986c7d758ecc3ad5ed6de1ebf2d1ae514a1fca9dee128244c4420a56b5c5066a5e
+    commit: 1ea16dd720f04cefa61f380ccc375d6f
+    digest: shake256:c51fd611a013929061384c1630d5e421a1083a62b6d5e4256c0a50b8abd4e8b91a944495918bbaa165d32398bbd6990a0c9b71f5275df9375386f2b050e67a6f
```
maximpertsov, line 13: did you mean to include this update?

martha-johnston: yeah I figured even tho the api I need is already included it might as well be up to date since I was adding another commit. can remove it tho if you think that's best

maximpertsov: I'm okay with that - just wanted to make sure it was intentional

PR #130, opened 2023-07-06
RSDK-3923 Add workflow to publish off-cycle releases

maximpertsov:
Allows off-cycle version to be published to the `next` distribution tag

TODO

* [ ] custom error message
* [ ] test


maximpertsov:
will reopen on upstream branch to test gh actions

PR #131, opened 2023-07-07
RSDK-3923 Add workflow to publish off-cycle versions

maximpertsov:
Allows off-cycle version to be published to the next distribution tag

### DIFF
--- a/.github/workflows/release_candidate.yml
+++ b/.github/workflows/release_candidate.yml
@@ -68,6 +68,8 @@ jobs:
         run: |
           if ${{ contains(steps.current_version.outputs.current_version, 'rc') }} ; then
             npm version prerelease --preid=rc --no-git-tag-version
+          elif ${{ contains(steps.current_version.outputs.current_version, 'next') && inputs.version == 'prepatch' }} ; then
+            npm version prerelease --preid=rc --no-git-tag-version
           else
             npm version ${{ inputs.version }} --preid=rc --no-git-tag-version
           fi
--- a/Makefile
+++ b/Makefile
@@ -71,6 +71,12 @@ build-js: $(node_modules) clean-js build-buf
 pack: build
 	npm pack
 
+# bump to the next pre-release version on the `next` distribution tag -
+# run this command if you need to publish an off-cycle release of the sdk
+.PHONY: bump-next-version
+bump-next-version:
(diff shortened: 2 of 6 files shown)


Makefile
```diff
@@ -71,6 +71,12 @@ build-js: $(node_modules) clean-js build-buf
 pack: build
 	npm pack
 
+# bump to the next pre-release version on the `next` distribution tag -
+# run this command if you need to publish an off-cycle release of the sdk
+.PHONY: bump-next-version
+bump-next-version:
+	npm version --no-git-tag-version prerelease --preid=next
```
maximpertsov, line 78: `next` versions will always be patch versions with `-next.X` suffixes. this might be a little awkward if next stable release ends up being a major or minor bump, since it will signal a patch bump that will next be released, but all in all not the worst situation

.github/workflows/release_candidate.yml
```diff
@@ -68,6 +68,8 @@ jobs:
         run: |
           if ${{ contains(steps.current_version.outputs.current_version, 'rc') }} ; then
             npm version prerelease --preid=rc --no-git-tag-version
+          elif ${{ contains(steps.current_version.outputs.current_version, 'next') && inputs.version == 'prepatch' }} ; then
+            npm version prerelease --preid=rc --no-git-tag-version
```
maximpertsov, line 72: if the current version is a `next` prerelease, running `prepatch` will increment the patch version one more - to avoid that, we run `prerelease --preid=rc` instead which will just change `-next` to `-rc`

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@viamrobotics/sdk",
-  "version": "0.2.2",
+  "version": "0.2.3-next.0",
```
maximpertsov, line 3: successfully published to `next` tag without adding to `latest` tag

cheukt (approved):
LGTM - I feel like a workflow that users can run to bump the version (to publish to npm) would also be useful, though then you'd have to make sure people aren't running it over and over

maximpertsov:
> LGTM - I feel like a workflow that users can run to bump the version (to publish to npm) would also be useful, though then you'd have to make sure people aren't running it over and over

Might add that in the future - for now I just want to bring back something similar to old workflow since it worked well for us.

PR #135, opened 2023-07-19
RSDK-2843 - grpc reconnection

purplenicole730:


src/robot/client.ts
```diff
@@ -114,12 +126,23 @@ export class RobotClient extends EventDispatcher implements Robot {
   constructor(
     serviceHost: string,
     webrtcOptions?: WebRTCOptions,
+    directOptions?: DirectOptions,
```
maximpertsov, line 129: let's make this the last option otherwise this break users who are currently creating a client via this constructor.

src/robot/client.ts
```diff
@@ -343,6 +373,7 @@ export class RobotClient extends EventDispatcher implements Robot {
         // eslint-disable-next-line no-await-in-loop
         await this.connecting;
       }
+      console.debug('reconnect while connecting - exiting early');
```
maximpertsov, line 376:
```suggestion
```

i think we can remove this log altogether

src/robot/groc-connection-manager.test.ts
maximpertsov, line 1: typo in filename

maximpertsov (approved):

PR #140, opened 2023-07-24
RSDK-3588 - add navigation service and moveonglobe

purplenicole730:


src/services/motion/motion.ts
```diff
@@ -1,4 +1,6 @@
+import { GeoObstacle } from '../../gen/common/v1/common_pb';
```
maximpertsov, line 1: instead of passing the raw proto, let's use the `AsObject` object - we should add the following definition to custom types (`src/types.ts`) and use that:

```ts
// src/types.ts

export type GeoObstactle = common.GeoObstactle.AsObject;
```

src/services/navigation/client.ts
maximpertsov, line 1: we should also add a `src/navigation.ts` file for exports like we do for motion

maximpertsov (changes requested):
nice - let a few comments. we also need to export the new navigation client and friends in `src/main.ts`

maximpertsov:
also, let's bump the version via `make bump-next-version` if you plan to upgrade RC

src/main.ts
```diff
@@ -304,6 +304,11 @@ export { type DataManager, DataManagerClient } from './services/data-manager';
  * @group Raw Protobufs
  */
```
maximpertsov, line 305: change the docstring to

```
/**
 * Raw Protobuf interfaces for a Navigation service.
 *
 * Generated with https://github.com/improbable-eng/grpc-web
 *
 * @deprecated Use {@link NavigationClient} instead.
 * @alpha
 * @group Raw Protobufs
 */
```

maximpertsov (changes requested):
nice! one doc-string change request and then good to merge

PR #141, opened 2023-07-27
RSDK-2379 Add getStream to stream api

maximpertsov:
### Description 

Add a `getStream` method that adds and returns a stream from a camera. Mostly ported from the React teleop example (thanks @mcous!)

### Questions

* [x] I hardcoded a 30 second timeout to prevent this function from hanging. Is that reasonable? Should this be configurable? **
  * We settled on 5 seconds

### Testing

I ran the React teleop example using the function:

image

### DIFF
--- a/src/extra/stream/client.test.ts
+++ b/src/extra/stream/client.test.ts
@@ -46,4 +46,16 @@ describe('StreamClient', () => {
       fakeStream
     );
   });
+  test('getStream fails when add stream fails', async () => {
+    const error = new Error('could not add stream');
+    StreamServiceClient.prototype.addStream = vi
+      .fn()
+      .mockImplementation((_req, _md, cb) => {
+        cb(error);
+      });
+
+    await expect(streamClient.getStream('fakecam')).rejects.toStrictEqual(
+      error
+    );
+  });
 });
--- a/src/extra/stream/client.test.ts
+++ b/src/extra/stream/client.test.ts
@@ -46,12 +46,16 @@ describe('StreamClient', () => {
         streamClient.emit('track', { streams: [fakeStream] });
       });
 
+    const addStream = vi.spyOn(streamClient, 'add');
     await expect(streamClient.getStream(fakeCamName)).resolves.toStrictEqual(
       fakeStream
     );
+    expect(addStream).toHaveBeenCalledOnce();
+    expect(addStream).toHaveBeenCalledWith(fakeCamName);
   });
 
   test('getStream fails when add stream fails', async () => {
+    const fakeCamName = 'fakecam';
     const error = new Error('could not add stream');
     StreamServiceClient.prototype.addStream = vi
       .fn()
@@ -59,21 +63,28 @@ describe('StreamClient', () => {
         cb(error);
       });
 
-    await expect(streamClient.getStream('fakecam')).rejects.toThrow(error);
+    const addStream = vi.spyOn(streamClient, 'add');
+    await expect(streamClient.getStream(fakeCamName)).rejects.toThrow(error);
+    expect(addStream).toHaveBeenCalledOnce();
+    expect(addStream).toHaveBeenCalledWith(fakeCamName);
   });
 
   test('getStream fails when timeout exceeded', async () => {
+    const fakeCamName = 'fakecam';
     StreamServiceClient.prototype.addStream = vi
       .fn()
       .mockImplementation((_req, _md, cb) => {
         cb(null, {});
       });
 
-    const promise = streamClient.getStream('fakecam');
+    const addStream = vi.spyOn(streamClient, 'add');
+    const promise = streamClient.getStream(fakeCamName);
     vi.runAllTimers();
     await expect(promise).rejects.toThrowError(
       'Did not receive a stream after 5000 ms'
     );
+    expect(addStream).toHaveBeenCalledOnce();
+    expect(addStream).toHaveBeenCalledWith(fakeCamName);
   });
 
   test('getStream can add the same stream twice', async () => {
@@ -86,11 +97,14 @@ describe('StreamClient', () => {
         streamClient.emit('track', { streams: [fakeStream] });
       });
 
+    const addStream = vi.spyOn(streamClient, 'add');
     await expect(streamClient.getStream(fakeCamName)).resolves.toStrictEqual(
       fakeStream
     );
     await expect(streamClient.getStream(fakeCamName)).resolves.toStrictEqual(
       fakeStream
     );
+    expect(addStream).toHaveBeenCalledTimes(2);
+    expect(addStream).toHaveBeenCalledWith(fakeCamName);
   });
 });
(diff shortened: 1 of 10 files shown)


src/extra/stream/client.ts
```diff
@@ -97,4 +97,39 @@ export class StreamClient extends EventDispatcher implements Stream {
       this.streams.delete(name);
     }
   }
+
+  private getStreamTimeout = 30_000;
```
maximpertsov, line 101: 30 seconds is completely arbitrary, but I think we should have SOME timeout to prevent this functions from hanging if a corresponding stream never arrives for some reason

cheukt: do you have a sense of how long streams usually take to return? 30s feels like a very long time if it wasn't gonna arrive anyway. (wondering if we can shorten the timeout)

cheukt (approved):
LGTM :)

src/extra/stream/client.ts
```diff
@@ -97,4 +97,39 @@ export class StreamClient extends EventDispatcher implements Stream {
...
+          this.off('track', handleTrack as (args: unknown) => void);
+          resolve(stream);
+        }
+      };
+
+      this.on('track', handleTrack as (args: unknown) => void);
+
+      setTimeout(() => {
+        this.off('track', handleTrack as (args: unknown) => void);
+        reject(new Error('timed out'));
```
mcous, line 127: Could this be a more descriptive error?

```js
`Did not receive a stream after ${this.getStreamTimeout} ms`
```

src/extra/stream/client.ts
mcous: [nit] naming a const `getStreamTimeout` is a little confusing, and also should it live up with other properties?

src/extra/stream/client.test.ts
```diff
@@ -1,22 +1,94 @@
...
     }));
+
+  test('getStream creates and returns a new stream', async () => {
+    const fakeCamName = 'fakecam';
+    const fakeStream = { id: fakeCamName };
+    StreamServiceClient.prototype.addStream = vi
+      .fn()
+      .mockImplementation((_req, _md, cb) => {
+        cb(null, {});
+        streamClient.emit('track', { streams: [fakeStream] });
```
mcous, line 46: Should this be wrapped in an `if` that checks the arguments to ensure `addStream` is called correctly? This `mockImplementation` is unconditional, so `getStream` could call `addStream` incorrectly and this test would still pass

This question applies to all subsequent tests that fake out `addStream`. #92 shows an alternative approach that using a [library I wrote](https://mike.cousins.io/articles/2023-06-30-better-stubs/) that prevents unconditional mocking by design

src/extra/stream/client.test.ts
```diff
@@ -1,22 +1,94 @@
...
         done();
       });
 
       events.emit('track', { mock: true });
     }));
+
+  test('getStream creates and returns a new stream', async () => {
+    const fakeCamName = 'fakecam';
+    const fakeStream = { id: fakeCamName };
+    StreamServiceClient.prototype.addStream = vi
```
mcous, line 42: Since the `StreamClient` interacts with multiple methods of the `StreamServiceClient`, partially mocking out `StreamServiceClient` by making a single one of its methods risks making these tests brittle.

I'd recommend fixing the `vi.mock` call above to point to the correct path. That way, the entire `StreamServiceClient` will be fake. From there, I think you'll be able to do something like:

```ts
StreamServiceClient
  .prototype
  .addStream
  .mockImplementation((_req, _md, cb) => {
  // ...
```

For TypeScript support, you have to wrap the fake thing in `vi.mocked`:

```ts
vi.mocked(StreamServiceClient.prototype.addStream)
  .mockImplementation((_req, _md, cb) => {
  // ...
```

src/extra/stream/client.test.ts
```diff
@@ -1,22 +1,94 @@
...
 import { StreamClient } from './client';
 
+let robotClient: RobotClient;
+let streamClient: StreamClient;
+
 describe('StreamClient', () => {
+  beforeEach(() => {
+    vi.useFakeTimers();
+
+    vi.mock('./robot/client');
```
mcous, line 16: What's this trying to mock out and why?

mcous (commented):
Cool! Got some testing suggestions:

- The module mocking and resetting is misconfigured, leaving a bunch of [partial mocks](https://github.com/testdouble/contributing-tests/wiki/Partial-Mock) in play
- Mocks in tests are configured unconditionally, so we're not testing that `StreamServiceClient` is called correctly, only that its results are used correctly

src/extra/stream/client.test.ts
maximpertsov: The stream client requires a robot client to initialize

src/extra/stream/client.test.ts
maximpertsov: ah good call, will update

src/extra/stream/client.test.ts
maximpertsov: to confirm, the entire module will be mocked once i fix the mock path [as you pointed out here](https://github.com/viamrobotics/viam-typescript-sdk/pull/141/files#r1277791295), right? this was my intention

src/extra/stream/client.test.ts
```diff
@@ -1,22 +1,94 @@
...
+
+    robotClient = new RobotClient('fakehost');
+    vi.mock('./gen/proto/stream/v1/stream_pb_service');
```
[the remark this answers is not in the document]
maximpertsov: > I recommend putting all calls to vi.mock directly after your import statements

to confirm, if mock outside of `beforeEach`, will the module remain mocked even if i call `resetAllMocks`?

src/extra/stream/client.ts
```diff
@@ -97,4 +97,41 @@ export class StreamClient extends EventDispatcher implements Stream {
       this.streams.delete(name);
     }
   }
+
+  private STREAM_TIMEOUT = 5000;
+
+  /**
+   * Get a stream by name from a StreamClient. Will time out if stream is not
+   * received within 30 seconds.
```
cheukt, line 105: 30s -> 5s

maximpertsov:
```suggestion
   * received within 5 seconds.
```

src/extra/stream/client.test.ts
maximpertsov: > For TypeScript support, you have to wrap the fake thing in vi.mocked

hm, when I used `vi.mocked` as you described i got a lint exception  `@typescript-eslint/unbound method` that I wasn't quite able to figure out

src/extra/stream/client.test.ts
```diff
@@ -1,22 +1,94 @@
...
+
+    streamClient = new StreamClient(robotClient);
+  });
+
+  afterEach(() => {
+    vi.clearAllMocks();
```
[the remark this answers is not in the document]
maximpertsov: added to `vite.config.ts`

PR #142, opened 2023-07-28
RSDK-4057 Add Power Sensor Component 

oliviamiller:
Adds client side code for power sensor component and unit tests.

### DIFF
--- a/src/components/power-sensor/client.test.ts
+++ b/src/components/power-sensor/client.test.ts
@@ -7,7 +7,7 @@ import { PowerSensorServiceClient } from '../../gen/component/powersensor/v1/pow
 import { PowerSensorClient } from './client';
 
 let sensor: PowerSensorClient;
-const testPower = 0.5
+const testPower = 0.5;
 const testVoltage = [1.5, true];
 const testCurrent = [1, true];
 
@@ -21,7 +21,7 @@ beforeEach(() => {
     .mockImplementation((_req, _md, cb) => {
       cb(null, {
         getVolts: () => 1.5,
-        getIsAc: () => true
+        getIsAc: () => true,
       });
     });
   PowerSensorServiceClient.prototype.getCurrent = vi
@@ -29,7 +29,7 @@ beforeEach(() => {
     .mockImplementation((_req, _md, cb) => {
       cb(null, {
         getAmperes: () => 1,
-        getIsAc: () => true
+        getIsAc: () => true,
       });
     });
 
@@ -37,7 +37,7 @@ beforeEach(() => {
     .fn()
     .mockImplementation((_req, _md, cb) => {
       cb(null, {
-        getWatts: () => 0.5
+        getWatts: () => 0.5,
       });
     });
 
--- a/src/components/power-sensor/client.test.ts
+++ b/src/components/power-sensor/client.test.ts
@@ -7,9 +7,9 @@ import { PowerSensorServiceClient } from '../../gen/component/powersensor/v1/pow
 import { PowerSensorClient } from './client';
 
 let sensor: PowerSensorClient;
-const testVoltage: (number | boolean)[] = [0.5, true];
-const testCurrent: (number | boolean)[] = [1, false];
-let testPower: 2;
+const testPower = { Watts: 0.5 };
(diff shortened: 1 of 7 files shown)


src/components/power-sensor/power-sensor.ts
```diff
@@ -0,0 +1,18 @@
...
+  current?: number;
+  power?: number;
+};
+
+/** Represents any sensor that reports voltage, current, and/or power */
+export interface PowerSensor extends Sensor {
+  /** Get Voltage in volts */
+  getVoltage(extra?: StructType): Promise<readonly [number, boolean]>;
+  /** Get Current in amps */
+  getCurrent(extra?: StructType): Promise<readonly [number, boolean]>;
```
maximpertsov, line 15: we should document that getVoltage and getCurrent return whether the current is AC or not as the second return value

src/components/power-sensor/client.ts
```diff
@@ -0,0 +1,110 @@
...
+    );
+    return response.getWatts();
+  }
+
+  async getReadings(extra = {}) {
+    const readingFunctions = {
+      voltage: this.getVoltage.bind(this),
+      current: this.getCurrent.bind(this),
+      power: this.getPower.bind(this),
+    };
```
maximpertsov, line 82: assuming voltage and current will always have the current flow type for a given power source, maybe the following is a better return type:

```ts
{
  isAc: boolean,
  voltage: number,
  current: number,
  power: number,
}
```

(and we can throw an error if the two `isAc` values ever differ).

maximpertsov (changes requested):
doc update + change to readings return value

src/components/power-sensor/client.ts
oliviamiller: yes they should always be same, in go we decided to just return one of the isAC values and not check if they are different since that would be a driver error.  I'll change to do the same in TS

src/components/power-sensor/client.test.ts
```diff
@@ -7,9 +7,9 @@ import { PowerSensorServiceClient } from '../../gen/component/powersensor/v1/pow
 import { PowerSensorClient } from './client';
 
 let sensor: PowerSensorClient;
-const testPower = { Watts: 0.5 };
-const testVoltage = [{ Volts: 1 }, { IsAc: true }];
-const testCurrent = [{ Amperes: 1 }, { IsAc: true }];
+const testPower = 0.5;
+const testVoltage = [1.5, true];
+const testCurrent = [1, true];
```
maximpertsov, line 12: [nit] might be a bit cleaner / DRYer if you make these test variables refer to the individual values and have your mock functions return them as well. i.e.

```suggestion
const testPower = 0.5;
const testVoltage = 1.5;
const testCurrent = 1;
const testIsAc = true
```

src/components/power-sensor/client.test.ts
```diff
@@ -20,44 +20,24 @@ beforeEach(() => {
...
-            Volts: 1,
-          };
-        },
-        getIsAc: () => {
-          return {
-            IsAc: true,
-          };
-        },
+        getVolts: () => 1.5,
+        getIsAc: () => true,
```
maximpertsov, line 24: e.g. if you agree with my suggestion above - same for the ones below
```suggestion
        getVolts: () => testVoltage,
        getIsAc: () => testIsAc,
```

src/components/power-sensor/client.ts
```diff
@@ -71,31 +71,38 @@ export class PowerSensorClient implements PowerSensor {
...
-      })()
-    );
-    const entries = await Promise.all(tasks);
-
-    return Object.fromEntries(entries.flat()) as PowerSensorReadings;
+    const readings: Record<string, any> = {};
+    try {
+      const ret = await this.getVoltage(extra);
+      readings['voltage'] = ret[0];
+      readings['isAc'] = ret[1];
```
maximpertsov, line 83: [nit] (same for current and power below)

```suggestion
      const [voltage, isAc] = await this.getVoltage(extra);
      readings.voltage = voltage;
      readings.isAc = isAc;
```

maximpertsov (approved):
looks good minus linting errors, also left some style suggestions

PR #143, opened 2023-07-31
RSDK-1825 - add slam wrapper

purplenicole730:


src/services/slam/slam.ts
```diff
@@ -0,0 +1,9 @@
+import type { Resource } from '../../types';
+import type { GetPositionResponse, Timestamp } from './types';
+
+export interface Slam extends Resource {
+  getPosition: () => Promise<GetPositionResponse>;
+  getPointCloudMap: () => Promise<string | Uint8Array>;
+  getInternalState: () => Promise<string | Uint8Array>;
```
maximpertsov, line 7: there should be a method on the protobuf response object like `response.get<FIELD_NAME>_asU8` (see camera component client for an example), that returns the response data as a `Uint8Array`. The more we can constrain the return type the better.

```suggestion
  getPointCloudMap: () => Promise<Uint8Array>;
  getInternalState: () => Promise<Uint8Array>;
```

src/services/slam/types.ts
```diff
@@ -0,0 +1,5 @@
+import tspb from 'google-protobuf/google/protobuf/timestamp_pb';
+import pb from '../../gen/service/slam/v1/slam_pb';
+
+export type Timestamp = tspb.Timestamp.AsObject;
```
maximpertsov, line 4: instead of exposing pb duration, we should convert this to a javascript date - i think the conversion for this is:

```ts
new Date(duration.getSeconds() * 1e3 + duration.getNanos() / 1e6)
```

maximpertsov (changes requested):
we can make some return types a bit more constrained and idiomatic i think

src/services/slam/slam.ts
purplenicole730: This sounds great. I was also hesitant about this at first but thought it would come up in the review (which it did!)

maximpertsov (approved):

PR #146, opened 2023-08-03
RSDK-4338 Only publish docs with release

maximpertsov:
Only update [ts.viam.dev](https://ts.viam.dev/) when we release a stable version, rather than on every merge


cheukt:
would this also publish on release candidates?

.github/workflows/release.yml
```diff
@@ -89,7 +89,9 @@ jobs:
         uses: actions/upload-artifact@v3
         with:
           name: dist
-          path: viamrobotics-sdk-*.tgz
+          path: |
+            viamrobotics-sdk-*.tgz
+            docs/dist
```
maximpertsov, line 94: trying to remember - do we need to upload artifacts for any reason other than to break up the prepare and release jobs? if not maybe it's easier to combine them and remove all these artifact steps...

maximpertsov:
> would this also publish on release candidates?

no - should publish docs there too?

.github/workflows/release.yml
cheukt: I believe so, I think it's still worth separating the steps out tho

cheukt:
na, sorry I misread

cheukt (approved):

PR #147, opened 2023-08-03
RSDK-1825-2-remove-slam-streaming-methods

purplenicole730:


src/services/slam/slam.ts
```diff
@@ -12,15 +12,6 @@ export interface Slam extends Resource {
    */
   getPosition: () => Promise<GetPositionResponse>;
```
maximpertsov, line 13: ah, can we change this type name actually?

```suggestion
  getPosition: () => Promise<SlamPosition>;
```

maximpertsov (approved):

PR #150, opened 2023-08-04
RSDK-1825-2 - fix slam wrapper

purplenicole730:


maximpertsov (changes requested):
Two changes:

* let's add a test for the two streaming methods, since we are doing more than just converting a value in these cases
* [deprecate the raw proto interface](https://github.com/viamrobotics/viam-typescript-sdk/pull/147/files#diff-4fab5baaca5c14d2de62d8d2fceef376ddddcc8e9509d86cfa5643f51b89ce3dR328-R346)

src/services/slam/client.test.ts
```diff
@@ -0,0 +1,154 @@
...
+
+  SLAMServiceClient.prototype.getPointCloudMap = vi
+    .fn()
+    .mockImplementation(() => testPcdStream);
+  SLAMServiceClient.prototype.getInternalState = vi
+    .fn()
+    .mockImplementation(() => testInternalStream);
+
+  slam = new SlamClient(new RobotClient('host'), 'test-slam');
+});
```
maximpertsov, line 60: [opt] it might be helpful to make two [describe blocks](https://vitest.dev/api/#describe) here, one for `getPointCloudMap` and one for `getInternalStream`. totally find to leave as is if it's annoying to re-organize.

maximpertsov (commented):
looks good - just some mock configuration

maximpertsov (approved):

PR #155, opened 2023-08-16
RSDK-4477 update move on globe wrapper

maximpertsov:
hotfix-ish - the source code on `main` does not build properly, even though buf api definitions _should_ be pinned to early version of the motion api. This PR updates our wrappers to use the new motion api so we can build release candidates.

### DIFF
--- a/src/services/motion/client.ts
+++ b/src/services/motion/client.ts
@@ -24,7 +24,12 @@ import type {
   Transform,
   WorldState,
 } from '../../types';
-import { type Constraints, encodeConstraints } from './types';
+import {
+  type Constraints,
+  encodeConstraints,
+  type MotionConfiguration,
+  encodeMotionConfiguration,
+} from './types';
 import type { Motion } from './motion';
 
 /**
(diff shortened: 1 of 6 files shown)


src/services/motion/types.ts
```diff
@@ -57,3 +59,20 @@ export const encodeConstraints = (obj: Constraints): pb.Constraints => {
...
+    obj.visionServicesList.map((x) => encodeResourceName(x))
+  );
+  result.setPositionPollingFrequencyHz(obj.positionPollingFrequencyHz);
+  result.setObstaclePollingFrequencyHz(obj.obstaclePollingFrequencyHz);
+  result.setPlanDeviationM(obj.planDeviationM);
+  result.setLinearMPerSec(obj.linearMPerSec);
+  result.setAngularDegsPerSec(obj.angularDegsPerSec);
+
+  return result;
+};
```
maximpertsov, line 78: annoying, all of these fields are required in the resulting protobuf object, even tho they are marked as `optional` in the [source proto file](https://github.com/viamrobotics/api/blob/037415ecb066649cdbc955e2667902dac57c0fbf/proto/viam/service/motion/v1/motion.proto#L84-L97)

src/services/motion/client.ts
```diff
@@ -109,13 +114,13 @@ export class MotionClient implements Motion {
...
     obstaclesList?: GeoObstacle[],
-    linearMetersPerSec?: number,
-    angularDegPerSec?: number,
+    motionConfig?: MotionConfiguration,
     extra = {}
   ) {
     const { service } = this;
 
     const request = new pb.MoveOnGlobeRequest();
+    request.setName(this.name);
```
maximpertsov, line 123: this is a bug unrelated to the motion api change

src/services/motion/types.ts
purplenicole730: That's weird. Why are they required if they're supposed to be optional?

purplenicole730 (approved):
LGTM!

src/services/motion/types.ts
maximpertsov: yeah I'm not sure... but we can resolve this later since make them optional will still be backwards compatible - for now i just want to get his building again

PR #161, opened 2023-08-30
RSDK-4658 - add viam client

purplenicole730:
[Jira Ticket](https://viam.atlassian.net/browse/RSDK-4658)

Changes:
* Added a ViamClient that a user can connect to
* Added a ViamTransport that is made with the accessToken of a robot
* Added a DataClient with TabularDataByFilter method
* New example that demonstrates how to connect to ViamClient and call DataClient method.

Note: Decided independently not to include unit tests in this PR since I'm not entirely sure of the scope of adding tests for the ViamClient (previous experience trying to write tests for the RobotClient proved to be a very involved task). Please feel free to request if it feels necessary for the scope of this ticket.

The data is displayed like so: 
image


src/app/viam-client.ts
```diff
@@ -0,0 +1,35 @@
...
+import { DataClient } from './data-client';
+
+export class ViamClient {
+  private serviceHost: string;
+  private dialOpts: DialOptions;
+  private transportFactory: grpc.TransportFactory | undefined;
+  public dataServiceClient: DataServiceClient | undefined;
+  public dataClient: DataClient | undefined;
+
+  constructor(serviceHost: string, dialOpts: DialOptions) {
```
cheukt, line 14: will the service host ever change? should we have a default instead of forcing users to always have to specify a service host?

cheukt: also add a comment on what service host should be

src/main.ts
```diff
@@ -73,6 +90,17 @@ export { type Board, BoardClient } from './components/board';
 export { default as cameraApi } from './gen/component/camera/v1/camera_pb';
 export { type Camera, type MimeType, CameraClient } from './components/camera';
 
+/**
+ * Raw Protobuf interfaces for a Camera component.
+ *
+ * Generated with https://github.com/improbable-eng/grpc-web
+ *
+ * @deprecated Use {@link CameraClient} instead.
```
cheukt, line 98: this comment is wrong

src/main.ts
```diff
@@ -25,6 +25,23 @@ export {
  */
 export { RobotClient as Client } from './robot';
 
+/**
+ * @deprecated Use {@link RobotClient} instead.
```
cheukt, line 29: why is this deprecated?

examples/connect-app/tsconfig.json
```diff
@@ -0,0 +1,23 @@
+{
```
cheukt, line 1: please add a readme to this folder around how to run the example

examples/connect-app/index.html
```diff
@@ -0,0 +1,13 @@
+<!doctype html>
```
cheukt, line 1: this example should be a bit more substantive - maybe print the data we received after pressing a button to retrieve data?

src/app/data-client.ts
```diff
@@ -0,0 +1,58 @@
...
+      // eslint-disable-next-line no-await-in-loop
+      const response = await promisify<
+        TabularDataByFilterRequest,
+        TabularDataByFilterResponse
+      >(service.tabularDataByFilter.bind(service), req);
+      const dataList = response.getDataList();
+      if (!dataList || dataList.length === 0) {
+        break;
+      }
+      dataArray.push(...dataList.map((data) => data.toObject()));
```
cheukt, line 52: it'd be nice to put the timestamps in as an actual date object

src/app/data-client.ts
```diff
@@ -0,0 +1,58 @@
...
+
+    let last = '';
+    const dataArray: TabularData.AsObject[] = [];
+
+    for (;;) {
+      const dataReq = new DataRequest();
+      if (filter) {
+        dataReq.setFilter(filter);
+      } else {
+        dataReq.setFilter(new Filter());
```
cheukt, line 34: how hard is it to create a new filter? should we have convenience methods like in the python sdk?

cheukt (changes requested):

src/app/data-client.ts
purplenicole730: It's not that hard to create a filter (there's an example of it in the example). It doesn't seem necessary to have such a method.

cheukt: how does one convert a Date object into a timestamp?

src/main.ts
purplenicole730: I was just following the RobotClient above.

src/app/viam-client.ts
purplenicole730: Good point. In Python we don't even have a default. They hard coded it to always be `app.viam.com:443`. I will make that change.

src/app/data-client.ts
purplenicole730: The Timestamp object has a `fromDate()` function.

examples/connect-app/src/main.ts
```diff
@@ -0,0 +1,45 @@
...
+  };
+  return new VIAM.ViamClient(dialOpts);
+}
+
+async function main() {
+  let client: VIAM.ViamClient;
+  try {
+    console.log('app is connecting...');
+    client = await connect();
+    await client.connect();
```
HipsterBrown, line 28: For sake of clarity, either the `connect` function defined above should either call the `connect` method on the new ViamClient instance before returning that client instance, or be renamed to `getViamClient` since it does not perform the connection task at the moment.

```ts
client = getViamClient();
await client.connect();
```

or

```ts
async function connect(): Promise<VIAM.ViamClient> {
  // all the other variable setup
  const client = new VIAM.ViamClient(dialOpts);
  await client.connect();
  return client;
}

// within main
client = await connect()
```

src/app/data-client.ts
```diff
@@ -0,0 +1,58 @@
...
+export class DataClient {
+  private client: DataServiceClient;
+
+  constructor(serviceHost: string, grpcOptions: RpcOptions) {
+    this.client = new DataServiceClient(serviceHost, grpcOptions);
+  }
+
+  private get service() {
+    return this.client;
+  }
```
HipsterBrown, line 21: Is the `service` getter required if the `client` private instance property is renamed to `service`?

```suggestion
  private service: DataServiceClient;

  constructor(serviceHost: string, grpcOptions: RpcOptions) {
    this.service = new DataServiceClient(serviceHost, grpcOptions);
  }
```

Since it's a private value, I'm curious about the need for a private getter that just returns the DataServiceClient instance.

src/main.ts
```diff
@@ -25,6 +25,23 @@ export {
  */
 export { RobotClient as Client } from './robot';
 
+/**
+ * @deprecated Use {@link ViamClient} instead.
```
HipsterBrown, line 29: Since this is above the export for the `ViamClient`, I don't it is accurate to mark it as deprecated and link to the ViamClient.

I believe with is a part of other TSDoc comments above protobuf exports that were previously used before dedicate classes were created in their place.

src/main.ts
```diff
@@ -25,6 +25,23 @@ export {
...
+ * @group Clients
+ */
+export { ViamClient } from './app/viam-client';
+
+/**
+ * Raw Protobuf interfaces for a Data service.
+ *
+ * Generated with https://github.com/improbable-eng/grpc-web
+ *
+ * @deprecated Use {@link DataClient} instead.
```
HipsterBrown, line 39: Is the `DataClient` exported from this package or just the `ViamClient`, which has a public `dataClient` property as an instance of the `DataClient`?

src/main.ts
```diff
@@ -73,6 +90,17 @@ export { type Board, BoardClient } from './components/board';
 export { default as cameraApi } from './gen/component/camera/v1/camera_pb';
 export { type Camera, type MimeType, CameraClient } from './components/camera';
 
+/**
+ * Raw Protobuf interfaces for a Camera component.
```
HipsterBrown, line 94: Is this accurate for the `DataClient` type export?

src/robot/dial.ts
```diff
@@ -185,3 +193,44 @@ export const createRobotClient = async (
...
 };
+
+export const createViamTransportFactory = async (
+  serviceHost: string,
+  dialOpts: DialOptions
+): Promise<grpc.TransportFactory> => {
+  if (!dialOpts.credentials) {
+    throw new Error(`credential cannot be none`);
+  } else if (dialOpts.credentials.type === 'robot-secret') {
+    throw new Error(`credential type cannot be 'robot secret'`);
```
HipsterBrown, line 204: Is there a place we can link to or more information to share to help consumers of this package learn the correct credential configuration?

src/robot/dial.ts
```diff
@@ -185,3 +193,44 @@ export const createRobotClient = async (
...
+  const authClient = new AuthServiceClient(serviceHost, {
+    transport: transportFactory,
+  });
+
+  const accessToken = await new Promise<string>((resolve, reject) => {
+    authClient.authenticate(req, new grpc.Metadata(), (err, response) => {
+      if (err) {
+        return reject(err);
+      }
+      const token = response?.getAccessToken().toString() ?? '';
```
HipsterBrown, line 228: Would an empty access token be considered an authentication error?

src/robot/dial.ts
maximpertsov: I don't think this is documented anywhere - let's add a doc-string to this function with these usage details.

src/app/viam-client.ts
cheukt: I would like to see the service host have a default, but continue to allow users to specify a different URL - this will be good for testing

examples/connect-app/src/main.ts
```diff
@@ -0,0 +1,45 @@
+import { type DialOptions } from '@viamrobotics/rpc/src/dial';
```
cheukt, line 1: now that org api keys are in, the app example should use api keys https://github.com/viamrobotics/viam-typescript-sdk/pull/164

cheukt (changes requested):

src/main.ts
purplenicole730: Nope! Haha, I fixed that :)

src/app/data-client.ts
maximpertsov: i actually don't remember why we started using this pattern... we don't need this extra function

src/robot/dial.ts
purplenicole730: According to @cheukt, it will throw an authentication error elsewhere, so there isn't a pressing need for there to be a check here.

examples/connect-app/src/main.ts
purplenicole730: While org api keys are now supported, we have not made strides to easily access the org api keys, per conversation with @jr22. To get the org api key, we would need to [create the key](https://docs.viam.com/manage/cli/#create-an-organization-api-key) by authenticating with an `accessToken`, which is what we're trying to get in the first place by using the key.

purplenicole730: tested with an actual api key, and ready to go!

src/robot/dial.ts
```diff
@@ -185,3 +193,50 @@ export const createRobotClient = async (
...
+ * credential type to use would preferably be the organization api key.
+ */
+export const createViamTransportFactory = async (
+  serviceHost: string,
+  dialOpts: DialOptions
+): Promise<grpc.TransportFactory> => {
+  if (!dialOpts.credentials) {
+    throw new Error(`credential cannot be none`);
+  } else if (dialOpts.credentials.type === 'robot-secret') {
+    throw new Error(`credential type cannot be 'robot secret'`);
```
HipsterBrown, line 210: A more informative error message could like look:

```suggestion
    throw new Error(`credential type cannot be 'robot-secret'. Must be either 'robot-location-secret' or 'api-key'.`);
```

src/robot/dial.ts
```diff
@@ -185,3 +193,50 @@ export const createRobotClient = async (
...
+  } else if (!dialOpts.authEntity) {
+    throw new Error(`auth entity cannot be None`);
```
HipsterBrown, line 212: `None` is not a value in JS/TS, so a more helpful message might be:

```suggestion
    throw new Error(`auth entity cannot be null, undefined, or an empty value.`);
```

HipsterBrown (approved):
LGTM

Nice work on this! 🎉

maximpertsov (approved):

cheukt (approved):
LGTM!

PR #164, opened 2023-09-15
APP-2520: Update typescript sdk to work with api keys

jr22:
Ticket: https://viam.atlassian.net/browse/APP-2520

Updated the dial code to support API keys. Here is a code sample for how connecting with api keys works: 
```
async function connect(): Promise<VIAM.RobotClient> {
  // You can remove this block entirely if your robot is not authenticated.
  // Otherwise, replace with an actual secret.
  const credential = {
    payload: "<API-KEY>",
    type: 'api-key',
  };

  // Replace with the host of your actual robot running Viam.
  const host = '<HOST>';

  // Replace with the signaling address. If you are running your robot on Viam,
  // it is most likely https://app.viam.com:443.
  const signalingAddress = '<SIGNALING_ADDRESS>';

  // You can replace this with a different ICE server, append additional ICE
  // servers, or omit entirely. This option is not strictly required but can
  // make it easier to connect via WebRTC.
  const iceServers = [{ urls: 'stun:global.stun.twilio.com:3478' }];

  return VIAM.createRobotClient({
    host,
    credential,
    authEntity: "<API-KEY-ID>",
    signalingAddress,
    iceServers,
    // optional: configure reconnection options
    reconnectMaxAttempts: 7,
    reconnectMaxWait: 1000,
  });
}

```
The biggest difference here is that the authEntity is actually the api key ID and not the host. I updated the code to conditionally use authEntity to connect if it exists and fall back on the old behavior of impliedURL otherwise. I manually tested that this works for both location secrets and API keys (although we are planning to deprecate secrets soon). 

Heres the [scope](https://docs.google.com/document/d/17PELjzsNhFz55SByTuA4WNiM9yyes3Rbp92G1c6ZMP0/edit#heading=h.tcicyojyqi6c) and [spec](https://docs.google.com/document/d/1ovx8AKwxobkO2860vLKHPpbKexNrmepYVXsoLYwEg60/edit#heading=h.tcicyojyqi6c) that describes the changes for org api keys. We had to make similar changes to the python, c++ and rust SDKs. 
* https://github.com/viamrobotics/viam-cpp-sdk/pull/144
* https://github.com/viamrobotics/rust-utils/pull/75
* https://github.com/viamrobotics/viam-python-sdk/pull/419

Let me know if there any tests I should update! Also happy to chat offline if you want more details.

### DIFF
--- a/src/robot/dial.ts
+++ b/src/robot/dial.ts
@@ -112,7 +112,7 @@ const dialWebRTC = async (conf: DialWebRTCConf): Promise<RobotClient> => {
   if (conf.credential) {
     creds = conf.credential;
   }
-  await client.connect(impliedURL, creds);
+  await client.connect(conf.authEntity || impliedURL, creds);
 
   // eslint-disable-next-line no-console
   console.debug('connected via WebRTC');


maximpertsov (commented):
We should also add the same logic for direct gRPC connections ([here](https://github.com/viamrobotics/viam-typescript-sdk/pull/164/files#diff-52a50b491f74e91af65fba14f575d59124ae522a5fb352e0185c7cfeedd3cd82L53))

You can test this by running same script you ran but omitting `signalingAddress` and `iceServers` from the options provided to `createRobotClient`:

```
  return VIAM.createRobotClient({
    host,
    credential,
    authEntity: "<API-KEY-ID>",
    // optional: configure reconnection options
    reconnectMaxAttempts: 7,
    reconnectMaxWait: 1000,
  });
```

maximpertsov (approved):

PR #165, opened 2023-09-26
RSDK-4439 - add more data methods

purplenicole730:


src/app/data-client.test.ts
```diff
@@ -0,0 +1,263 @@
...
+      const partId = 'testPartId';
+      const locationsIdsList = ['testLocationId1', 'testLocationId2'];
+      const organizationIdsList = ['testOrgId1', 'testOrgId2'];
+      const mimeTypeList = ['testMimeType1', 'testMimeType2'];
+      const bboxLabelsList = ['testBboxLabel1', 'testBboxLabel2'];
+      const startTime = new Date(1, 1, 1, 1, 1, 1);
+      const endTime = new Date(2, 2, 2, 2, 2, 2);
+      const interval = new CaptureInterval();
+      interval.setStart(Timestamp.fromDate(startTime));
+      interval.setEnd(Timestamp.fromDate(endTime));
```
maximpertsov, line 221: let's make it so the `createFilter` function accepts `CaptureInterval.asObject`, so it can be defined as a plain object, e.g.

```suggestion
      const interval = { start: startTime, end: endTime };
```

src/app/data-client.test.ts
```diff
@@ -0,0 +1,263 @@
...
+      const tagsList = ['testTag1', 'testTag2'];
+      const tagsFilter = new TagsFilter();
+      tagsFilter.setTagsList(tagsList);
```
maximpertsov, line 224: same comment as [here](https://github.com/viamrobotics/viam-typescript-sdk/pull/165/files#r1338964065). Let's update the input type so users can define the tagsFilter as:

```suggestion
      const tagsFilter = ['testTag1', 'testTag2'];
```

src/app/data-client.test.ts
```diff
@@ -0,0 +1,263 @@
...
+      dataReq.setFilter(filter);
+      dataReq.setLimit(100);
+      dataReq.setLast('');
+      const req = new TabularDataByFilterRequest();
+      req.setDataRequest(dataReq);
+      req.setCountOnly(false);
+
+      await subject().tabularDataByFilter(filter);
+      expect(methodSpy).toHaveBeenCalledWith(
+        req,
```
HipsterBrown, line 83: How do you feel about a more verbose name for the expected request to match the internal call to the data service client?

``suggestion
      const expectedRequest = new TabularDataByFilterRequest();
      expectedRequest.setDataRequest(dataReq);
      expectedRequest.setCountOnly(false);

      await subject().tabularDataByFilter(filter);
      expect(methodSpy).toHaveBeenCalledWith(
        expectedRequest,
```

File: src/app/data-client.test.ts (hunk 1 above)
Comment (maximpertsov) [reply src/app/data-client.test.ts:221]: ugh, ignore this! this is already the case - i mixed up the test with the expections

File: src/app/data-client.test.ts (hunk 2 above)
Comment (maximpertsov) [reply src/app/data-client.test.ts:224]: ugh, ignore this! this is already the case - i mixed up the test with the expections

File: src/app/data-client.test.ts [hunk 4]
```diff
@@ -0,0 +1,263 @@
...
+        mimeTypeList,
+        bboxLabelsList,
+        startTime,
+        endTime,
+        tags: tagsList,
+      };
+      const testFilter = subject().createFilter(opts);
+      expect(testFilter.getComponentType()).toEqual('testComponentType');
+
+      const actualFilter = new Filter();
```
maximpertsov, line 245: [minor] 

```suggestion
      const actualFilter = subject().createFilter(opts);
      expect(actualFilter.getComponentType()).toEqual('testComponentType');

      const expectedFilter = new Filter();
```

src/app/data-client.ts
```diff
@@ -67,6 +69,65 @@ export class DataClient {
...
+
+    for (;;) {
+      const dataReq = new pb.DataRequest();
+      if (filter) {
+        dataReq.setFilter(filter);
+      } else {
+        dataReq.setFilter(new pb.Filter());
+      }
+      dataReq.setLimit(100);
+      dataReq.setLast(last);
```
HipsterBrown, line 86: I know this is how the loop is set up in the `tabularDataByFilter` method, however I'm wonder if a new `DataRequest` needs to be created for each iteration. The filter and limit don't change between requests, just the `last` value. 

Could this work and slightly reduce the amount of work in each iteration?

```suggestion
    const dataReq = new pb.DataRequest();
    dataReq.setFilter(filter ?? new pb.Filter());
    dataReq.setLimit(100);
    
    for (;;) {
      dataReq.setLast(last);
```

src/app/data-client.test.ts
```diff
@@ -0,0 +1,263 @@
...
+    });
+
+    test('create filter', () => {
+      const opts = { componentName: 'camera' };
+      const testFilter = subject().createFilter(opts);
+
+      const actualFilter = new Filter();
+      actualFilter.setComponentName('camera');
+
+      expect(testFilter).toEqual(actualFilter);
```
maximpertsov, line 202: [minor] we should rename "actualFilter" to something like "expectedFilter" - I'm used to "actual" referring to the result of the subject function that we defined, not the the value we say it should be equal to.

Same goes for all of the tests below if you agree.

```suggestion
      const testFilter = subject().createFilter(opts);

      const expectedFilter = new Filter();
      expectedFilter.setComponentName('camera');

      expect(testFilter).toEqual(expectedFilter);
```

maximpertsov (commented):
looks good! just one suggestion on how some test variables are named

src/app/data-client.ts
```diff
@@ -67,6 +69,65 @@ export class DataClient {
...
+    const { service } = this;
+
+    const binaryIds: pb.BinaryID[] = [];
+    for (const id of ids) {
+      const binaryId = new pb.BinaryID();
+      binaryId.setFileId(id.fileId);
+      binaryId.setOrganizationId(id.organizationId);
+      binaryId.setLocationId(id.locationId);
+      binaryIds.push(binaryId);
+    }
```
HipsterBrown, line 118: Similar to how the tabular data response was mapped into the data list, the `binaryIds` list can be created by mapping over the list of `ids`:

```suggestion
    const binaryIds: pb.BinaryID[] = ids.map(({ fileId, organizationId, locationId }) => {
      const binaryId = new pb.BinaryID();
      binaryId.setFileId(fileId);
      binaryId.setOrganizationId(organizationId);
      binaryId.setLocationId(locationId);
      return binaryId;
    });
```

HipsterBrown (commented):
Overall, this looks really good and nice to have the initial test coverage in place. Just a few suggestions but nothing blocking, just seeing what you think.

src/app/data-client.test.ts
purplenicole730: I was just following how it was in the data client. I don't really have an opinion on this, so I'll make this change

maximpertsov (approved):

HipsterBrown (approved):
LGTM 🎉

PR #166, opened 2023-09-28
RSDK-5032 - strip authentity of protocol string

purplenicole730:
Connection to the robot fails if the `authEntity` has a protocol string (i.e. "https://") in the beginning.

The solution was to grab the host of the `authEntity` string in ``createRobotClient`.

Tested by replacing the `main` of the vanilla example with the code sample provided from `app.viam.com` for a robot. Locally connect to a laptop as a robot through gRPC.

### DIFF
--- a/src/robot/dial.ts
+++ b/src/robot/dial.ts
@@ -156,6 +156,7 @@ export const createRobotClient = async (
   conf: DialConf
 ): Promise<RobotClient> => {
   let client;
+  conf.authEntity = conf.authEntity?.replace(/^(?<http>.*:\/\/)/u, '');
 
   if (conf.reconnectMaxAttempts && !isPosInt(conf.reconnectMaxAttempts)) {
     throw new Error(
--- a/src/robot/dial.ts
+++ b/src/robot/dial.ts
@@ -156,7 +156,9 @@ export const createRobotClient = async (
   conf: DialConf
 ): Promise<RobotClient> => {
   let client;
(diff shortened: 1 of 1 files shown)


src/robot/dial.ts
```diff
@@ -156,6 +156,7 @@ export const createRobotClient = async (
   conf: DialConf
 ): Promise<RobotClient> => {
   let client;
+  conf.authEntity = conf.authEntity?.replace(/^(?<http>.*:\/\/)/u, '');
```
maximpertsov, line 159: let's use the built-in URL parser to get the host

```suggestion
  if (conf.authEntity) {
    conf.authEntity = new URL(conf.authEntity).host;
  }
```

maximpertsov (changes requested):
in addition to the suggestion, can you leave a note about how this was tested?

purplenicole730:
> in addition to the suggestion, can you leave a note about how this was tested?

@maximpertsov What do you mean? Do you want me to say that the change was tested by replacing the vanilla `main`?

maximpertsov:
> > in addition to the suggestion, can you leave a note about how this was tested?
> 
> @maximpertsov What do you mean? Do you want me to say that the change was tested by replacing the vanilla `main` with my own laptop as the local robot?

yeah, any details about the scripts your ran, the setup, etc would be helpful

maximpertsov (approved):

PR #180, opened 2023-10-18
RSDK-5362 generate protos from pinned versions

maximpertsov:
Add a custom script that pulls our current pinned proto version from `buf.lock`. Update Makefile to generate using those pinned versions.


scripts/get-buf-lock-version.js
maximpertsov, line 1: I made this a JS script instead of a bash script to avoid installing dependencies outside of node (we needed a yaml parser in this case).

doc/dependency_decisions.yml
```diff
@@ -89,3 +89,9 @@
     :license_links: https://github.com/yinzara/protoc-gen-js/blob/master/LICENSE
     :versions: []
     :when: 2023-05-12 14:28:29.173104997 Z
+- - :permit
+  - ISC
+  - :who:
+    :why:
+    :versions: []
+    :when: 2023-10-18 20:31:08.255446206 Z
```
maximpertsov, line 97: [permitted in another repo](https://github.com/viamrobotics/goutils/blob/429830ca0acd683d214b1ed926e64569e1a507ac/doc/dependency_decisions.yml#L27-L31) - similar to MIT / simplified BSD

cheukt (approved):

PR #181, opened 2023-10-24
RSDK-5228 - allow setting sdp values

purplenicole730:
[Related goutils change](https://github.com/viamrobotics/goutils/pull/215)
[Jira ticket](https://viam.atlassian.net/browse/RSDK-5228)

Allowing for a custom priority attribute to be pushed in.

Tested with a board running the MicroRDK to confirm default and customized priority works as expected. Also tested with a robot running the regular RDK (with and without a priority attribute) to confirm everything else still works.

NOTE: The `build_lint_test` error seems to be caused by goutils, and should be solved with the goutils change linked above.

How could we test this, or is it okay not to add tests for this?

### DIFF
--- a/src/robot/client.ts
+++ b/src/robot/client.ts
@@ -414,7 +414,8 @@ export class RobotClient extends EventDispatcher implements Robot {
         },
       };
 
-      if (priority && opts.webrtcOptions) {
+      // Webrtcoptions will always be defined, but TS doesn't know this
+      if ((priority === 0 || priority) && opts.webrtcOptions) {
         opts.webrtcOptions.additionalSdpFields = { 'x-priority': priority };
       }
 
--- a/src/robot/client.ts
+++ b/src/robot/client.ts
@@ -372,7 +372,7 @@ export class RobotClient extends EventDispatcher implements Robot {
   public async connect(
     authEntity = this.savedAuthEntity,
     creds = this.savedCreds,
-    priority = 0
+    priority?: number
   ) {
     if (this.connecting) {
       // This lint is clearly wrong due to how the event loop works such that after an await, the condition may no longer be true.
--- a/src/robot/client.ts
+++ b/src/robot/client.ts
@@ -370,9 +370,9 @@ export class RobotClient extends EventDispatcher implements Robot {
   }
 
   public async connect(
+    priority?: number,
     authEntity = this.savedAuthEntity,
-    creds = this.savedCreds,
-    priority?: number
+    creds = this.savedCreds
   ) {
     if (this.connecting) {
       // This lint is clearly wrong due to how the event loop works such that after an await, the condition may no longer be true.
@@ -405,7 +405,7 @@ export class RobotClient extends EventDispatcher implements Robot {
         webrtcOptions: {
           disableTrickleICE: false,
           rtcConfig: this.webrtcOptions?.rtcConfig,
-          priority: priority,
+          priority,
         },
       };
 
--- a/src/robot/client.ts
+++ b/src/robot/client.ts
@@ -56,9 +56,9 @@ interface SessionOptions {
 }
 
 export interface ConnectOptions {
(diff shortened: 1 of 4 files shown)


maximpertsov:
> How could we test this, or is it okay not to add tests for this?

I think it's okay - I'm more concerned with the goutils side of the change since the JS code has zero test coverage. It seems like you did sufficient manual testing there though. It might be a good idea to run the [echo example](https://github.com/viamrobotics/goutils/tree/main/rpc/examples/echo) in that repo to ensure all of the different gRPC over webrtc calls still work.

src/robot/client.ts
```diff
@@ -370,6 +370,7 @@ export class RobotClient extends EventDispatcher implements Robot {
   }
 
   public async connect(
+    priority?: number,
     authEntity = this.savedAuthEntity,
     creds = this.savedCreds
   ) {
```
DTCurrie, line 376: Since we are updating the arguments here, could we update `connect` to accept an `object` of options? Should making updating this signature (especially with optional properties) much safer/easier moving forward.

```suggestion
  export interface ConnectOptions {
    priority?: number;
    authEntity?: string;
    creds?: Credentials; 
  }

  public async connect({
    priority,
    authEntity = this.savedAuthEntity,
    creds = this.savedCreds,
  }: ConnectOptions) {
```

cc: @micheal-parks

purplenicole730: I actually really like this idea. Thank you for bringing it up! I noticed this as well and wished I could do this, but wasn't sure if it was a good idea to change such an important function. I'll be making this change now that I know others agree with me!

src/robot/client.ts
```diff
@@ -369,11 +375,11 @@ export class RobotClient extends EventDispatcher implements Robot {
...
 
-  public async connect(
-    priority?: number,
+  public async connect({
+    priority,
     authEntity = this.savedAuthEntity,
-    creds = this.savedCreds
-  ) {
+    creds = this.savedCreds,
+  }: ConnectOptions) {
```
micheal-parks, line 382: I'd recommend making this object optional so that `this.connect()` could still be called with no args:

```ts
  public async connect({
    priority,
    authEntity = this.savedAuthEntity,
    creds = this.savedCreds,
  }: ConnectOptions = {}) {
```

purplenicole730: Thanks for that suggestion! Made

maximpertsov (approved):

src/robot/client.ts
```diff
@@ -407,6 +414,10 @@ export class RobotClient extends EventDispatcher implements Robot {
         },
       };
 
+      if (priority && opts.webrtcOptions) {
```
cheukt, line 417: isn't opts.webrtcOptions going to always evaluate to true? if so we should just remove it as part of the if statement

src/robot/dial.ts
```diff
@@ -120,7 +121,11 @@ const dialWebRTC = async (conf: DialWebRTCConf): Promise<RobotClient> => {
   if (conf.credential) {
     creds = conf.credential;
   }
-  await client.connect(conf.authEntity || impliedURL, creds);
+  await client.connect({
+    priority: conf.priority,
```
cheukt, line 125: nit: move this to the bottom

cheukt (approved):
lgtm with a question

src/robot/client.ts
cheukt, line 417: also what if priority is 0? that should still be passed through but would evaluate to false here

cheukt (changes requested):

src/robot/client.ts
purplenicole730: Yes, when you read the code, but for some reason Typescript says that it could be undefined

cheukt (approved):

DTCurrie (approved):
✅

PR #185, opened 2023-10-27
RSDK-5477 - clearer tabular data objects

purplenicole730:
Originally, the data object just came with a metadata index and printed out unclear data. Now the data object includes the metadata for each data object. We now also call `toJavaScript()` to display the data, which makes things clearer.
The data now looks like this:
```  
  {
    "Lat": demo,
    "Lng": demo,
    "metadata": {
      "organizationId": "demo",
      "locationId": "demo",
      "robotName": "demo",
      "robotId": "demo",
      "partName": "demo",
      "partId": "demo",
      "componentType": "demo",
      "componentName": "demo",
      "methodName": "demo",
      "methodParametersMap": [],
      "tagsList": [],
      "mimeType": "demo"
    },
    "timeRequested": "2023-10-27T19:06:00.497Z",
    "timeReceived": "2023-10-27T19:06:00.497Z"
  },
```

### DIFF
--- a/src/app/data-client.ts
+++ b/src/app/data-client.ts
@@ -30,7 +30,7 @@ export class DataClient {
   /**
    * Filter and download tabular data. The returned metadata might be empty if
    * the metadata index of the data is out of the bounds of the returned
-   * metadata list
+   * metadata list.
    *
    * @param filter - Optional `pb.Filter` specifying tabular data to retrieve.
    *   No `filter` implies all tabular data.
--- a/src/app/data-client.ts
+++ b/src/app/data-client.ts
@@ -54,16 +54,16 @@ export class DataClient {
       }
       dataArray.push(
         ...dataList.map((data) => {
-          const index = data.getMetadataIndex();
+          const mdIndex = data.getMetadataIndex();
           const mdListLength = response.getMetadataList().length;
-          if (mdListLength != 0 && index >= mdListLength) {
+          if (mdListLength != 0 && mdIndex >= mdListLength) {
             throw Error(
-              `metadata index ${index} is out of response's metadata list`
+              `metadata index ${mdIndex} is out of response's metadata list`
             );
           }
           return {
             ...data.getData()?.toJavaScript(),
-            metadata: response.getMetadataList()[index]?.toObject(),
+            metadata: response.getMetadataList()[mdIndex]?.toObject(),
             timeRequested: data.getTimeRequested()?.toDate(),
             timeReceived: data.getTimeReceived()?.toDate(),
           };
--- a/src/app/data-client.ts
+++ b/src/app/data-client.ts
@@ -14,7 +14,7 @@ export type FilterOptions = Partial<pb.Filter.AsObject> & {
 };
 
 type TabularData = {
-  data?: googleStructPb.JavaScriptValue;
+  data?: { [key: string]: googleStructPb.JavaScriptValue };
   metadata?: pb.CaptureMetadata.AsObject;
   timeRequested?: Date;
   timeReceived?: Date;
@@ -28,6 +28,10 @@ export class DataClient {
   }
 
   async tabularDataByFilter(filter?: pb.Filter) {
+    /*
+     * The returned metadata might be empty if the metadata index of
+     * the data is out of the bounds of the returned metadata list
+     */
     const { service } = this;
 
     let last = '';
@@ -57,10 +61,6 @@ export class DataClient {
       dataArray.push(
         ...dataList.map((data) => {
           const mdIndex = data.getMetadataIndex();
-          /*
-           * The returned metadata might be empty if the metadata index of
-           * the data is out of the bounds of the returned metadata list
-           */
           const metadata =
             mdListLength !== 0 && mdIndex >= mdListLength
               ? new pb.CaptureMetadata().toObject()
--- a/src/app/data-client.ts
+++ b/src/app/data-client.ts
@@ -15,7 +15,7 @@ export type FilterOptions = Partial<pb.Filter.AsObject> & {
 
 type TabularData = {
   data?: googleStructPb.Struct.AsObject;
-  metadataIndex: number;
+  metadata?: pb.CaptureMetadata.AsObject;
   timeRequested?: Date;
   timeReceived?: Date;
 };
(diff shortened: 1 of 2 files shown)


felixreichenbach (commented):
Am I correct with the assumption that we transfer the metadata only once  (like a list of meta data entries) parallel to the data?

Are we going to provide a merged list (metadata added to each data object) or is this up to the developer to merge accordingly?

I think both options are valuable depending on the use of the data.

purplenicole730:
> Am I correct with the assumption that we transfer the metadata only once (like a list of meta data entries) parallel to the data?
> 
> Are we going to provide a merged list (metadata added to each data object) or is this up to the developer to merge accordingly?
> 
> I think both options are valuable depending on the use of the data.

Yes, the metadata list comes all at once. If you look at the example in the PR comment, you'll see the metadata already merged in with the array that comes out of calling `tabularDataByFilter()`. I have confirmed and tested that the metadata is included in the binary functions.

src/app/data-client.ts
```diff
@@ -15,7 +15,7 @@ export type FilterOptions = Partial<pb.Filter.AsObject> & {
 
 type TabularData = {
   data?: googleStructPb.Struct.AsObject;
```
cheukt, line 17: type here probably needs to change as well

felixreichenbach:
> > Am I correct with the assumption that we transfer the metadata only once (like a list of meta data entries) parallel to the data?
> > Are we going to provide a merged list (metadata added to each data object) or is this up to the developer to merge accordingly?
> > I think both options are valuable depending on the use of the data.
> 
> @felixreichenbach Yes, the metadata list comes all at once. If you look at the example in the PR comment, you'll see the metadata already merged in with the array that comes out of calling `tabularDataByFilter()`. I have confirmed and tested that the metadata is included in the binary functions.
> 
> Are you suggesting that we allow the developers to toggle this on and off?

I think most people want it merged like the example. I wouldn't add anything more for now.

src/app/data-client.ts
```diff
@@ -53,11 +53,21 @@ export class DataClient {
...
+        ...dataList.map((data) => {
+          const mdIndex = data.getMetadataIndex();
+          const mdListLength = response.getMetadataList().length;
+          if (mdListLength != 0 && mdIndex >= mdListLength) {
+            throw Error(
+              `metadata index ${mdIndex} is out of response's metadata list`
+            );
+          }
+          return {
+            ...data.getData()?.toJavaScript(),
```
cheukt, line 65: can we have `data: data.getData()?.toJavaScript()`? otherwise the data is just exploded into this object and not sure if that's what we want

src/app/data-client.ts
```diff
@@ -53,11 +53,21 @@ export class DataClient {
...
       }
       dataArray.push(
-        ...dataList.map((data) => ({
-          ...data.toObject(),
-          timeRequested: data.getTimeRequested()?.toDate(),
-          timeReceived: data.getTimeReceived()?.toDate(),
-        }))
+        ...dataList.map((data) => {
+          const mdIndex = data.getMetadataIndex();
+          const mdListLength = response.getMetadataList().length;
```
cheukt, line 58: we can move this line before the loop starts, since this number will always be the same

src/app/data-client.ts
```diff
@@ -53,11 +53,21 @@ export class DataClient {
...
+          if (mdListLength != 0 && mdIndex >= mdListLength) {
+            throw Error(
```
cheukt: I wouldn't error, would just have metadata be blank instead (and document that metadata might be blank in the method docstring) - the data is likely still useful in that case

cheukt (changes requested):
some changes requested, also can you rename the pr to `RSDK-5477 - Clearer tabular data objects`

src/app/data-client.ts
```diff
@@ -52,12 +52,26 @@ export class DataClient {
...
-          ...data.toObject(),
-          timeRequested: data.getTimeRequested()?.toDate(),
-          timeReceived: data.getTimeReceived()?.toDate(),
-        }))
+        ...dataList.map((data) => {
+          const mdIndex = data.getMetadataIndex();
+          /*
+           * The returned metadata might be empty if the metadata index of
+           * the data is out of the bounds of the returned metadata list
+           */
```
cheukt, line 63: this should be at the top of the method like https://github.com/viamrobotics/viam-typescript-sdk/blob/a886344f35c1593647c120a110fb266ceb356eb8/src/types.ts#L12-L16, so it'll show up in ts.viam.dev

src/app/data-client.ts
```diff
@@ -14,8 +14,8 @@ export type FilterOptions = Partial<pb.Filter.AsObject> & {
 };
 
 type TabularData = {
-  data?: googleStructPb.Struct.AsObject;
-  metadataIndex: number;
+  data?: googleStructPb.JavaScriptValue;
```
cheukt, line 17: I think this will actually be a `{[key: string]: googleStructPb.JavaScriptValue};` based on what toJavaScript() returns

cheukt (changes requested):
more small nits

src/app/data-client.ts
purplenicole730: I wonder why it passed as just a regular javascriptvalue type. Is it okay that this was not picked up by lint?

cheukt: that's weird, but maybe just a weird issue

src/app/data-client.ts
```diff
@@ -28,6 +28,10 @@ export class DataClient {
   }
 
   async tabularDataByFilter(filter?: pb.Filter) {
+    /*
+     * The returned metadata might be empty if the metadata index of
+     * the data is out of the bounds of the returned metadata list
+     */
```
cheukt, line 34:
```suggestion
    /**
     * Filter and download tabular data. The returned metadata might be empty if the
     * metadata index of the data is out of the bounds of the returned metadata list
     * @param filter - Optional `pb.Filter` specifying tabular data to retrieve. No `filter` implies all tabular data.
     **/
  async tabularDataByFilter(filter?: pb.Filter) {
```

cheukt (changes requested):
the comment needs to go above the method name

cheukt (approved):
LGTM!

PR #191, opened 2023-11-03
RSDK-4790: Add Readings to TypeScript SDK

martha-johnston:
add getreadings to movement and power sensor

### DIFF
--- a/src/components/movement-sensor/client.test.ts
+++ b/src/components/movement-sensor/client.test.ts
@@ -9,12 +9,14 @@ import { MovementSensorClient } from './client';
 let sensor: MovementSensorClient;
 
 interface Val {
-  toJavaScript(): string
-} 
+  toJavaScript(): string;
+}
 
 const mapVals: Val = {
-  toJavaScript: () => { return 'readings'}
-}
+  toJavaScript: () => {
+    return 'readings';
+  },
+};
 
 beforeEach(() => {
   RobotClient.prototype.createServiceClient = vi
@@ -24,11 +26,13 @@ beforeEach(() => {
   MovementSensorServiceClient.prototype.getReadings = vi
     .fn()
     .mockImplementation((_req, _md, cb) => {
-      cb(null, { getReadingsMap: () => {
-        return {
-          entries: () =>  new Map<string, unknown>([['readings', mapVals]])
-        }
-      }});
+      cb(null, {
+        getReadingsMap: () => {
+          return {
+            entries: () => new Map<string, unknown>([['readings', mapVals]]),
+          };
+        },
+      });
     });
 
   MovementSensorServiceClient.prototype.getPosition = vi
@@ -123,6 +127,6 @@ test('get readings returns without unimplemented fields', async () => {
     unimplementedError
   );
   await expect(sensor.getReadings()).resolves.toStrictEqual({
-    readings: 'readings'
+    readings: 'readings',
   });
 });
--- a/src/components/movement-sensor/client.test.ts
+++ b/src/components/movement-sensor/client.test.ts
@@ -1,6 +1,7 @@
 // @vitest-environment happy-dom
 
 import { afterEach, beforeEach, expect, test, vi } from 'vitest';
+import { Value } from 'google-protobuf/google/protobuf/struct_pb'
 
 import { RobotClient } from '../../robot';
 import { MovementSensorServiceClient } from '../../gen/component/movementsensor/v1/movementsensor_pb_service';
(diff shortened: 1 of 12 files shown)


src/components/movement-sensor/client.test.ts
```diff
@@ -8,11 +8,33 @@ import { MovementSensorClient } from './client';
 
 let sensor: MovementSensorClient;
 
+interface Val {
+  toJavaScript(): string;
+}
+
+const mapVals: Val = {
```
purplenicole730, line 15: I'm realizing now that there probably doesn't have to be an interface here, and you could just create `mapVals` directly without a type.

purplenicole730, line 15:
```suggestion
const mapVals: = {
```

purplenicole730: I'm realizing that we don't need a type here. We could probably just directly create mapVals

src/components/power-sensor/client.test.ts
```diff
@@ -12,6 +12,16 @@ const testVoltage = 1.5;
 const testCurrent = 1;
 const testIsAc = true;
 
+interface Val {
+  toJavaScript(): string;
+}
+
+const mapVals: Val = {
```
purplenicole730, line 19:
```suggestion
const mapVals = {

File: src/main.ts [hunk 3]
```diff
@@ -166,11 +166,7 @@ export {
  * @group Raw Protobufs
  */
 export { default as powerSensorApi } from './gen/component/powersensor/v1/powersensor_pb';
-export {
-  type PowerSensor,
-  type PowerSensorReadings,
-  PowerSensorClient,
-} from './components/powersensor';
+export { type PowerSensor, PowerSensorClient } from './components/powersensor';
```
purplenicole730, line 169: Does this change not have to happen with `MovementSensorReadings` as well?

purplenicole730 (changes requested):
Looks mostly good, but `MovementSensorReadings` should be completely removed, and not exported by `main.ts`.

src/components/movement-sensor/client.test.ts
martha-johnston: good point, I'll remove it

purplenicole730 (approved):
LGTM!

src/components/power-sensor/client.ts
```diff
@@ -76,29 +80,23 @@ export class PowerSensorClient implements PowerSensor {
...
+      powersensorService.getReadings.bind(powersensorService),
+      request
+    );
+
+    const result: Record<string, unknown> = {};
+    for (const [key, value] of response.getReadingsMap().entries()) {
+      result[key] = value.toJavaScript();
     }
-    return readings as PowerSensorReadings;
+    return result;
```
DTCurrie, line 99: Why are we using `Promise<Record<string, unknown>>` here instead of the `PowerSensorReadings` type?

martha-johnston: because I thought PowerSensorReadings was a type we were using before there was an API for powersensor.GetReadings

DTCurrie: We relied on those types in the RC, so now we see these errors:

```/__w/rdk/rdk/web/frontend/src/components/power-sensor/index.svelte:27:5
Error: Type 'unknown' is not assignable to type 'number | undefined'. (ts)
    const readings = await powerSensorClient.getReadings();
    voltageValue = readings.voltage;
    currentValue = readings.current;


/__w/rdk/rdk/web/frontend/src/components/power-sensor/index.svelte:28:5
Error: Type 'unknown' is not assignable to type 'number | undefined'. (ts)
    voltageValue = readings.voltage;
    currentValue = readings.current;
    powerValue = readings.power;


/__w/rdk/rdk/web/frontend/src/components/power-sensor/index.svelte:29:5
Error: Type 'unknown' is not assignable to type 'number | undefined'. (ts)
    currentValue = readings.current;
    powerValue = readings.power;
  } catch (error) {
  ```
  
With these changes we can no longer rely on the SDK to return us `current`, `voltage`, or `power` as a part of the readings. With this change we no longer know if those fields exist on the response object, and since they are type to `unknown` we have no idea if they are numbers, strings, or objects.

martha-johnston: oh sorry I had no idea, are you seeing this issue with movement sensor as well? or is it still working

DTCurrie: I haven't even tested anything out, the RC is failing it's build checks right now with an SDK version bump. These changes are not passing our standard checks.

PR #194, opened 2023-11-10
RSDK-4790 FIX: export powersensor from robot/client.ts

martha-johnston:
fully just forgot this file which is causing issues now that I'm trying to make the frontend changes


maximpertsov (approved):
Ah, we need to add something to CI to catch this sort of thing.

PR #198, opened 2023-11-13
Add default check connection interval to be 10 seconds

njooma:
The SDK didn't actually set the `heartbeatIntervalMs` for the gRPC check connection interval. This creates a default value of 10 seconds.

### DIFF
--- a/src/robot/grpc-connection-manager.ts
+++ b/src/robot/grpc-connection-manager.ts
@@ -16,17 +16,18 @@ const timeoutBlob = new Blob(
 export default class GRPCConnectionManager {
   private innerTransportFactory: grpc.TransportFactory;
   private client: RobotServiceClient;
-  private heartbeatIntervalMs: number | undefined;
+  private heartbeatIntervalMs: number;
 
   public connecting: Promise<void> | undefined;
   private connectResolve: (() => void) | undefined;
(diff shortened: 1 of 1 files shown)


maximpertsov (approved):

maximpertsov (changes requested):
~actually, it's more consistent to use `backoff` like we do for webrtc connections: https://github.com/viamrobotics/viam-typescript-sdk/blob/25d52c3580c272f9ceed31b69f49a7ee316fb806/src/robot/client.ts#L174~

Ignore, this is unrelated - this change is fine.

maximpertsov (approved):

njooma:
I'm not sure if this is a `backoff` situation -- this isn't retrying to establish a connection that might not exist, but rather the interval for checking that the connection does exist (which I think should be at a set interval rather than backing off). Unless I'm missing something?

maximpertsov:
> I'm not sure if this is a `backoff` situation -- this isn't retrying to establish a connection that might not exist, but rather the interval for checking that the connection does exist (which I think should be at a set interval rather than backing off). Unless I'm missing something?

Yeah it's totally unrelated, sorry about that.

stuqdog (approved):

PR #201, opened 2023-11-20
Bump @viamrobotics/rpc to 0.1.39

ethanlook:
Bumping the `@viamrobotics/rpc` version so it is fixed as a transitive dependency in `app`.

See my PR for the downstream issue: https://github.com/viamrobotics/app/actions/runs/6895151896/job/18758485918?pr=2866

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "0.7.0",
       "license": "Apache-2.0",
       "dependencies": {
-        "@viamrobotics/rpc": "^0.1.38",
+        "@viamrobotics/rpc": "^0.1.39",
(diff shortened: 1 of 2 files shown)


maximpertsov (approved):
@ethanlook just checking - do you need a new version of the typescript SDK with this updated dependency to be available now? or can it be released in the normal release cycle?

ethanlook:
> @ethanlook just checking - do you need a new version of the typescript SDK with this updated dependency to be available now? or can it be released in the normal release cycle?

@maximpertsov thanks for asking - I can be patient and wait for the normal release cycle! I can work around in `app` in the meantime.

PR #202, opened 2023-11-22
RSDK-5735 Create app client with access token

maximpertsov:
Changes:

* App client accepts access tokens that are obtained from an outside flows (like oauth for example).

Flyby:

* Require the app clients to be created with the `createViamClient` - this breaks the current interface, but will allow us to make similar interfaces in a non-breaking way in the future.

### DIFF
--- a/examples/connect-app/src/main.ts
+++ b/examples/connect-app/src/main.ts
@@ -1,19 +1,18 @@
-import { type DialOptions } from '@viamrobotics/rpc/src/dial';
 import * as VIAM from '@viamrobotics/sdk';
 
+const API_KEY_ID = import.meta.env.VITE_API_KEY_ID;
+const API_KEY_SECRET = import.meta.env.VITE_API_KEY_SECRET;
+
 async function connect(): Promise<VIAM.ViamClient> {
-  const credential = {
-    payload: '<API-KEY>',
-    type: 'api-key',
+  const opts: VIAM.ViamClientOptions = {
+    credential: {
+      type: 'api-key',
+      authEntity: API_KEY_ID,
+      payload: API_KEY_SECRET,
+    },
   };
 
-  const dialOpts: DialOptions = {
-    authEntity: '<API-KEY-ID>',
-    credentials: credential,
-  };
-
-  const client = new VIAM.ViamClient(dialOpts);
-  await client.connect();
+  const client = await VIAM.createViamClient(opts);
 
   return client;
 }
@@ -22,6 +21,7 @@ const button = <HTMLButtonElement>document.getElementById('main-button');
 
 async function run(client: VIAM.ViamClient) {
   // A filter is an optional tool to filter out which data comes back.
+  // const opts: VIAM.FilterOptions = { componentType: 'camera' };
   const opts: VIAM.FilterOptions = { componentType: 'camera' };
   const filter = client.dataClient.createFilter(opts);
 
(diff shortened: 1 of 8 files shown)


src/app/viam-transport.ts
maximpertsov, line 1: The diff looks slightly bigger than what actually changed since this logic was also moved from [./src/robot/dial](https://github.com/viamrobotics/viam-typescript-sdk/pull/202/files#diff-52a50b491f74e91af65fba14f575d59124ae522a5fb352e0185c7cfeedd3cd82). 

The main functional change is that users can now provide either an access token for direct access or a credential to be exchanged for an access token. Only the latter was accepted before.

src/main.ts
```diff
@@ -26,7 +26,17 @@ export {
  */
 export { RobotClient as Client } from './robot';
 
-export { ViamClient } from './app/viam-client';
+export {
+  createViamClient,
+  type ViamClient,
```
maximpertsov, line 31: This is a breaking change - users won't be able to instantiate a `ViamClient` with the class constructor anymore and must use `createViamClient` instead.

This is more idiomatic and will make it easier to make non-breaking changes to interface moving forward.

We should do this for other class constructors as well but that's a thing for another day.

src/app/viam-transport.ts
maximpertsov, line 1: The diff looks slightly bigger than what actually changed since this logic was also moved from [./src/robot/dial.ts](https://github.com/viamrobotics/viam-typescript-sdk/pull/202/files#diff-52a50b491f74e91af65fba14f575d59124ae522a5fb352e0185c7cfeedd3cd82). 

The main functional change is that users can now provide either an access token for direct access or a credential to be exchanged for an access token. Only the latter was accepted before.

src/app/viam-transport.ts
```diff
@@ -1,4 +1,91 @@
 import { grpc } from '@improbable-eng/grpc-web';
```
maximpertsov, line 1: The diff looks slightly bigger than what actually changed since this logic was also moved from [./src/robot/dial.ts](https://github.com/viamrobotics/viam-typescript-sdk/pull/202/files#diff-52a50b491f74e91af65fba14f575d59124ae522a5fb352e0185c7cfeedd3cd82). 

The main functional change is that users can now provide either an access token for direct access or a credential to be exchanged for an access token. Only the latter was accepted before.

examples/connect-app/src/main.ts
```diff
@@ -1,36 +1,35 @@
...
-  // A filter is an optional tool to filter out which data comes back.
-  const opts: VIAM.FilterOptions = { componentType: 'camera' };
-  const filter = client.dataClient.createFilter(opts);
-
   try {
     button.disabled = true;
     const textElement = <HTMLParagraphElement>document.getElementById('text');
     textElement.innerHTML = 'waiting for data...';
 
-    const dataList = await client.dataClient.tabularDataByFilter(filter);
```
maximpertsov, line 33: This potentially returns a lot of data - updating to use `tabularDataBySQL` with a hard limit.

njooma (approved):
So I guess the point is two-fold:

1. We don't need the user to provide dial options because there are no options that are configurable except the credentials, so might as well simplify the DX here
2. This removes the need to call `client.connect` to actually make the call, further improcing the DX

LGTM, but you should post in the breaking changes channel to make people aware

PR #204, opened 2023-11-27
[RSDK-5872] MQL Query BSON

njooma:
Convert MQL queries to BSON rather than strings


dmhilly (approved):
This LGTM! Is it ok that this is a breaking change/do we need to be backwards compatible for a while? (The API will be backwards compatible at least until the frontend is updated to supply bytes rather than a string)

njooma:
@dmhilly I think we rip the bandaid and make this a breaking change directly -- the usage is probably minimal at this point since we only released it yesterday

PR #209, opened 2023-12-07
RSDK-4825: Add getPaths to navigation service

ethanlook:
I need `navService.getPaths` in order to render paths in the navigation service RC card - I figured the easiest thing was just to quickly add it myself, but let me know if there's a better way to go about this!

## Change log

- Add `getPaths` to service client implementation and interface
- Add `Path` type to types definition
- `make bump-next-version`

## Review requests

- It looks like the pattern is to not add tests for simple pass-through RPCs. Is there a test I should add?
- Will the `next` version automatically get published to NPM?

### DIFF
--- a/src/services/navigation/types.ts
+++ b/src/services/navigation/types.ts
@@ -3,3 +3,4 @@ import pb from '../../gen/service/navigation/v1/navigation_pb';
 export type ModeMap = pb.ModeMap;
 export type Waypoint = pb.Waypoint.AsObject;
 export type NavigationPosition = pb.GetLocationResponse.AsObject;
+export type Path = pb.Path.AsObject;
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@viamrobotics/sdk",
-  "version": "0.8.0",
+  "version": "0.8.1-next.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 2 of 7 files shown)


src/services/navigation/types.ts
```diff
@@ -3,3 +3,4 @@ import pb from '../../gen/service/navigation/v1/navigation_pb';
 export type ModeMap = pb.ModeMap;
 export type Waypoint = pb.Waypoint.AsObject;
 export type NavigationPosition = pb.GetLocationResponse.AsObject;
+export type Path = pb.Path.AsObject;
```
maximpertsov, line 6: also export this type [in src/main.ts](https://github.com/viamrobotics/viam-typescript-sdk/blob/b6c61ff0a030fff17f279ecd875f1ac28109b92a/src/main.ts#L336-L341) as well if you want users to be able to import it. if that's not important than disregard

maximpertsov (commented):
> It looks like the pattern is to not add tests for simple pass-through RPCs. Is there a test I should add?

It's a judgment call - if the wrapper is just doing type conversations than it's probably fine to rely on `tsc` catching errors of that nature. This wrapper seems small enough where it's okay to omit.

> Will the next version automatically get published to NPM?

Yes it will

src/services/navigation/types.ts
ethanlook: Good catch - added!

maximpertsov (approved):

PR #211, opened 2023-12-11
RSDK-6016 - Add motion Expose Paths project RPC methods

nicksanford:
[Ticket](https://viam.atlassian.net/browse/RSDK-6016)

### Changes:

- Add `moveOnGlobeNew`
- Add `stopPlan`
- Add `getPlan`
- Add `listPlanStatuses`
- Add tests for above

Follow up Breaking Change: https://github.com/viamrobotics/viam-typescript-sdk/pull/212

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -320,7 +320,7 @@ export {
   type MotionConfiguration,
   type ObstacleDetector,
   type OrientationConstraint,
-  type PlanStateMap,
+  PlanState,
   MotionClient,
 } from './services/motion';
 
--- a/src/main.ts
+++ b/src/main.ts
@@ -311,13 +311,16 @@ export { default as inputControllerApi } from './gen/component/inputcontroller/v
  */
 export { default as motionApi } from './gen/service/motion/v1/motion_pb';
 export {
-  type Motion,
+  type CollisionSpecification,
   type Constraints,
+  type GetPlanResponse,
   type LinearConstraint,
+  type ListPlanStatusesResponse,
+  type Motion,
+  type MotionConfiguration,
   type ObstacleDetector,
   type OrientationConstraint,
-  type CollisionSpecification,
-  type MotionConfiguration,
+  type PlanStateMap,
   MotionClient,
 } from './services/motion';
 
--- a/src/main.ts
+++ b/src/main.ts
@@ -320,7 +320,7 @@ export {
   type MotionConfiguration,
   type ObstacleDetector,
   type OrientationConstraint,
-  PlanState,
+  type PlanState,
   MotionClient,
 } from './services/motion';
 
--- a/src/services/motion.ts
+++ b/src/services/motion.ts
@@ -6,5 +6,8 @@ export type {
   OrientationConstraint,
   CollisionSpecification,
   MotionConfiguration,
+  GetPlanResponse,
+  ListPlanStatusesResponse,
+  PlanStateMap,
 } from './motion/types';
 export { MotionClient } from './motion/client';
--- a/src/services/motion.ts
+++ b/src/services/motion.ts
@@ -8,6 +8,6 @@ export type {
   MotionConfiguration,
   GetPlanResponse,
   ListPlanStatusesResponse,
-  PlanStateMap,
+  PlanState,
 } from './motion/types';
 export { MotionClient } from './motion/client';
--- a/src/services/motion/client.test.ts
+++ b/src/services/motion/client.test.ts
@@ -1,0 +1,429 @@
+/*
+ *  @vitest-environment happy-dom
+ */
+
+import { type Mock, beforeEach, describe, expect, test, vi } from 'vitest';
+import { Struct } from 'google-protobuf/google/protobuf/struct_pb';
+import { Timestamp } from 'google-protobuf/google/protobuf/timestamp_pb';
+import { MotionServiceClient } from '../../gen/service/motion/v1/motion_pb_service';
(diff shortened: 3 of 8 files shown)


src/utils.ts
```diff
@@ -116,6 +116,10 @@ export const encodeResourceName = (
 const encodeGeometry = (obj: common.Geometry.AsObject): common.Geometry => {
   const result = new common.Geometry();
 
+  if (obj.label) {
+    result.setLabel(obj.label);
```
nicksanford, line 120: I found that this was not being set during testing. Yay tests!

src/services/motion/client.test.ts
```diff
@@ -0,0 +1,509 @@
...
+                  oY: 50,
+                  oZ: 60,
+                  theta: 70,
+                },
+              },
+            ],
+          ],
+        },
+        status: {
+          state: 1,
```
nicksanford, line 366: @maximpertsov how do I specify that this value is `motion_pb.PlanStateMap.PLAN_STATE_IN_PROGRESS`?

nicksanford:
@maximpertsov Is there a convenient way to test this against viam-server given RDK remote control doesn't use any of these apis?

maximpertsov:
> @maximpertsov Is there a convenient way to test this against viam-server given RDK remote control doesn't use any of these apis?

The most straightforward way would be to use the [example app](https://github.com/viamrobotics/viam-typescript-sdk/tree/main/examples/vanilla). You can either test against a local RDK server with a fake motion service or against a real robot on Viam. 

This app has one button that [runs some abritrary code](https://github.com/viamrobotics/viam-typescript-sdk/blob/c8e466076507dc31f23b205a59666f2f509581f4/examples/vanilla/src/main.ts#L42-L56) - you can replace that with any logic you need to test (i.e. a fake motion service). Just replace all of the placeholder `<...>` strings with appropriate values.

With a real robot, you can replace the contents of [src/main.ts](https://github.com/viamrobotics/viam-typescript-sdk/blob/c8e466076507dc31f23b205a59666f2f509581f4/examples/vanilla/src/main.ts) with the sample code from that robot and run the app and inspect the console logs.

maximpertsov (commented):
looks good, some minor things

src/services/motion/client.test.ts
```diff
@@ -0,0 +1,429 @@
...
+  let executionId: Mock<[], string>;
+
+  test('return executionID', async () => {
+    executionId = vi.fn(() => testExecutionId);
+
+    const expectedMotionName = motionClientName;
+    const expectedDestination = { latitude: 1, longitude: 2 };
+    const expectedObstaclesList: Array<pb.ObstacleDetector> = [];
+    const expectedHeading = 0;
+    const expectedComponenetName = {
```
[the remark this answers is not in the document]
nicksanford: done

src/services/motion/types.ts
```diff
@@ -7,6 +7,9 @@ export type LinearConstraint = pb.LinearConstraint.AsObject;
 export type OrientationConstraint = pb.OrientationConstraint.AsObject;
 export type CollisionSpecification = pb.CollisionSpecification.AsObject;
 export type MotionConfiguration = pb.MotionConfiguration.AsObject;
+export type GetPlanResponse = pb.GetPlanResponse.AsObject;
+export type ListPlanStatusesResponse = pb.ListPlanStatusesResponse.AsObject;
+export type PlanStateMap = pb.PlanStateMap;
```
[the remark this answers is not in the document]
nicksanford: done

src/services/motion/client.test.ts
nicksanford: done

src/services/motion/types.ts
maximpertsov, line 12: Instead of exporting the PlanStateMap, let's export this type:

```suggestion
type ValueOf<T> = T[keyof T];
export const { PlanState } = pb;
export type PlanState = ValueOf<typeof pb.PlanState>;
```

We do this for [another enum type](https://github.com/viamrobotics/viam-typescript-sdk/blob/c8e466076507dc31f23b205a59666f2f509581f4/src/components/encoder/encoder.ts#L6-L8). This allows library users to use `PlanState` as either enum value lookup or type of value.

For example:

```
import { PlanState } as VIAM from '@viamrobotics/sdk';

// as a value lookup
PlanState.PLAN_STATE_FAILED // => 4

// as a type
function foo(ps: PlanState) { ... } // ps must be 0,1,2,3,4 or compilation fails
```

src/services/motion.ts
```diff
@@ -6,5 +6,8 @@ export type {
   OrientationConstraint,
   CollisionSpecification,
   MotionConfiguration,
+  GetPlanResponse,
+  ListPlanStatusesResponse,
+  PlanStateMap,
```
maximpertsov, line 11:
```suggestion
  PlanState,
```

src/main.ts
```diff
@@ -311,13 +311,16 @@ export { default as inputControllerApi } from './gen/component/inputcontroller/v
...
+  type GetPlanResponse,
   type LinearConstraint,
+  type ListPlanStatusesResponse,
+  type Motion,
+  type MotionConfiguration,
   type ObstacleDetector,
   type OrientationConstraint,
-  type CollisionSpecification,
-  type MotionConfiguration,
+  type PlanStateMap,
```
maximpertsov, line 323:
```suggestion
  PlanState,
```

maximpertsov (approved):

PR #221, opened 2023-12-21
[RSDK-6167] Update RPC library

njooma:
The package `@viamrobotics/rpc` was updated to a new version that now uses Vite. However, that introduced issues in the existing SDK. This PR address those issues:

* No longer imports from non-standard `/src/dial` file
* Updates the `@viamrobotics/eslint-config` package to keep up to date with the rest of our TS offerings
  * This involved updated many other outdated packages
  * This also involved fixing/disabling new ESLint errors that popped up

This requires the merge of https://github.com/viamrobotics/goutils/pull/227, followed by a new goutils release, after which we can reinstall/test the new RPC library

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,8 +31,7 @@ test-watch: $(node_modules) build-buf
 lint: $(node_modules) build-buf
 	npm run lint
 	npm run typecheck
-	# TODO(RSDK-5407): We can stop ignoring `@viamrobotics/rpc` once build issues are resolved in the latest version.
-	npm run check -- -i @viamrobotics/rpc
+	npm run check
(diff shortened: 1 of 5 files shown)


mcous (commented):
Can we split this into two PRs? This will be easier to review and test if we do the ESLint tooling bump first, then follow up with the RPC bump

njooma:
@mcous Added a new PR that only does dependency updates for ESLint and other tooling: https://github.com/viamrobotics/viam-typescript-sdk/pull/223

mcous (approved):
Code changes LGTM

PR #222, opened 2023-12-26
RSDK-6033 Add GetProperties to Nav and SLAM

jeremyrhyde:
This PR adds the GetProperties endpoints for both the slam and navigation services in viam's typescript sdk. The new endpoint was added in [this](https://github.com/viamrobotics/api/commit/223673cee35c340ea2fb40b742680c5104abdafc) commit to the API.

**JIRA Ticket:** https://viam.atlassian.net/browse/RSDK-6033

### DIFF
--- a/src/services/navigation.ts
+++ b/src/services/navigation.ts
@@ -4,5 +4,6 @@ export type {
   Waypoint,
   NavigationPosition,
   Path,
+  NavigationProperties,
 } from './navigation/types';
 export { NavigationClient } from './navigation/client';
--- a/src/services/slam.ts
+++ b/src/services/slam.ts
@@ -1,3 +1,3 @@
 export type { Slam } from './slam/slam';
-export type { SlamPosition } from './slam/types';
+export type { SlamPosition, SlamProperties } from './slam/types';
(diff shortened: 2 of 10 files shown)


src/services/navigation.ts
```diff
@@ -4,5 +4,6 @@ export type {
   Waypoint,
   NavigationPosition,
   Path,
+  NavigationProperties,
```
purplenicole730, line 7: If we're exporting NavigationProperties, we should also export this in `src/main.ts`.

src/services/slam.ts
```diff
@@ -1,3 +1,3 @@
 export type { Slam } from './slam/slam';
-export type { SlamPosition } from './slam/types';
+export type { SlamPosition, SlamProperties } from './slam/types';
```
purplenicole730, line 2: Same as NavProperties, should export this in `src/main.ts`.

purplenicole730 (approved):
If we want users to have access to the type `NavigationProperties` and `SlamProperties`, then we should probably export to `main.ts`.

Otherwise, looks good to me!

maximpertsov (approved):

PR #223, opened 2023-12-28
Update deps

njooma:
This is the sister PR for https://github.com/viamrobotics/viam-typescript-sdk/pull/221, with package updates for tooling


doc/dependency_decisions.yml
```diff
@@ -58,29 +58,37 @@
...
     :why: 'TODO(RSDK-583): why is license finder unable to detect this license automatically?'
     :license_links: https://github.com/viamrobotics/goutils/blob/main/LICENSE
     :versions: []
     :when: 2023-05-01 18:19:05.224492180 Z
 - - :license
-  - "@viamrobotics/typescript-config"
+  - '@viamrobotics/typescript-config'
   - Apache 2.0
   - :who:
     :why: 'TODO(RSDK-583): why is license finder unable to detect this license automatically?'
```
mcous, line 80: I don't know what this file is, but I know the answer to this question and how to fix it, if you want to assign this ticket to me!

src/app/viam-client.test.ts
```diff
@@ -29,7 +29,7 @@ describe('ViamClient', () => {
     payload: 'testAccessToken',
   };
 
-  const subject = () => createViamClient(options!);
+  const subject = async () => createViamClient(options!); // eslint-disable-line @typescript-eslint/no-non-null-assertion
```
mcous, line 32: This is probably better as an override for `*.test.ts` files in the eslint configuration rather than a one-off disable comment.

Alternatively, if the project is reorganized a little to follow our test file conventions from other FE projects, the base config will automatically disable this rule in all tests.

- `src`
  - `some-dir`
    - `some-module.ts`
    - `__tests__`
      - `some-module.spec.ts`

src/events.ts
```diff
@@ -8,6 +8,7 @@ export class EventDispatcher {
 
   on(type: string, listener: Callback) {
     const { listeners } = this;
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
```
mcous, line 11: This seems like potentially an actual issue with the type annotations. We may want to set these failing rules to `warn` in the eslint config for future follow up rather than disabling with comments, which is the general strategy we take with linting upgrades that surface new errors

src/responses.ts
```diff
@@ -2,16 +2,17 @@ import { EventDispatcher } from './events';
 import type { ResponseStream } from './gen/robot/v1/robot_pb_service';
 
 export class ViamResponseStream<T> extends EventDispatcher {
-  private stream: ResponseStream<any>;
+  private stream: ResponseStream<any>; // eslint-disable-line @typescript-eslint/no-explicit-any
```
mcous, line 5: Same comment, we should set this to warn so we follow up and fix the typing

src/responses.ts
```diff
@@ -2,16 +2,17 @@ import { EventDispatcher } from './events';
...
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   constructor(stream: ResponseStream<any>) {
     super();
     this.stream = stream;
   }
 
   override on(
     type: string,
-    handler: (message: any) => void
+    handler: (message: any) => void // eslint-disable-line @typescript-eslint/no-explicit-any
```
mcous, line 15: Warn rather than suppress

src/responses.ts
```diff
@@ -2,16 +2,17 @@ import { EventDispatcher } from './events';
 import type { ResponseStream } from './gen/robot/v1/robot_pb_service';
 
 export class ViamResponseStream<T> extends EventDispatcher {
-  private stream: ResponseStream<any>;
+  private stream: ResponseStream<any>; // eslint-disable-line @typescript-eslint/no-explicit-any
 
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
mcous, line 7: Warn rather than suppress

src/robot/client.ts
```diff
@@ -741,6 +744,7 @@ export class RobotClient extends EventDispatcher implements Robot {
     request.setEvery(new Duration().setNanos(durationMs * 1e6));
 
     const statusStream = robotService.streamStatus(request);
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
```
mcous, line 747: Warn rather than suppress, unless this is to prevent an auto fix, in which case can you add a TODO comment to go along with the suppression?

src/robot/session-manager.ts
```diff
@@ -141,6 +141,7 @@ export default class SessionManager {
         new grpc.Metadata(),
         (err, resp) => {
           if (err) {
+            // eslint-disable-next-line @typescript-eslint/no-unsafe-enum-comparison
```
mcous, line 144: Warn rather than suppress

src/robot/session-transport.ts
```diff
@@ -20,7 +20,7 @@ export default class SessionTransport implements grpc.Transport {
       if (
         err &&
         err instanceof GRPCError &&
-        err.code === grpc.Code.InvalidArgument &&
+        err.code === grpc.Code.InvalidArgument && // eslint-disable-line @typescript-eslint/no-unsafe-enum-comparison
```
mcous, line 23: Warn rather than suppress

mcous (commented):
I think most of these ESLint suppression comments should be warnings in the config, instead, for later follow up.

In the app and Prime, we switched over to using `__tests__` directories for in-src tests, and the base configs understand to disable certain rules (like "no any") in test directories, which looks like it could help here, too

Makefile
```diff
@@ -31,7 +31,7 @@ test-watch: $(node_modules) build-buf
 lint: $(node_modules) build-buf
 	npm run lint
 	npm run typecheck
-        // TODO(RSDK-5407): We can stop ignoring `@viamrobotics/rpc` once build issues are resolved in the latest version.
+	# TODO(RSDK-5407): We can stop ignoring `@viamrobotics/rpc` once build issues are resolved in the latest version.
```
mcous, line 34: Be careful about putting comments inside the body of Makefile recipes; they will be sent to the shell. It'll usually be ok, because a # signifies a comment in bash, too, but every now and again you can really weird issues from it (or at least confusing CI logs). Generally, I try to keep comments outside of recipes and put them above the target, instead

src/events.ts
njooma: I think that's a good strategy overall and will implement, but for this line in particular shouldn't it be ignored? 

My understanding of this line is that `listeners` is a map of `String: Set`. Since the key represented by the var `type` might not be in the map, it could come back as undefined, in which case we do a conditional assignment to a new `Set` (while not assigning if it exists). 

I don't see this line as an issue, but maybe I'm missing something?

mcous: I think this is complaining because `listeners` is defined as a `Record<string, Set<...>>`, which says "every string field is a `Set`." This lint is telling you "if every string field is a set, then you don't need to conditionally assign." I think this this issue can be fixed by setting the type signature to say "every string field is `Set` or is `undefined`" using `Partial`

```ts
listeners: Partial<Record<string, Set<Callback>>> = {};
```

package.json
```diff
@@ -53,28 +53,28 @@
...
+    "@typescript-eslint/parser": "^6.15.0",
+    "@viamrobotics/eslint-config": "^0.3.0",
+    "@viamrobotics/prettier-config": "^0.3.4",
+    "@viamrobotics/typescript-config": "^0.1.0",
     "concurrently": "^8.0.1",
     "copyfiles": "^2.4.1",
     "eslint": "^8.37.0",
-    "eslint-config-prettier": "^8.5.0",
+    "eslint-config-prettier": "^9.1.0",
     "eslint-plugin-tsdoc": "^0.2.17",
```
mcous, line 65: I don't see this plugin included in `.eslintrc.cjs`; is it used?

Makefile
njooma: Good call. As this is temporary (will get removed when we merge in the PR to update RPC), I'll leave it for now. But good practice

package.json
njooma: Not sure -- I removed it and it didn't seem to cause any problems so.. I'll keep it out

doc/dependency_decisions.yml
njooma: This file let's us manually approve libraries. I'll assign the ticket to you for now! I think it's more broad than just JS stuff, but if you could tackle the JS bit (and maybe explain why?), we might be able to fix the other SDK issues ourselves

mcous: Nice! For the JS side, it's a rather trivial packaging issue that switching to pnpm workspaces resolves

purplenicole730 (approved):
From my limited knowledge about this, LGTM

package.json
mcous: It looks like you removed `eslint-config-prettier` (which is used as a peer dep of `@viamrobotics/eslint-config`) instead of `eslint-plugin-tsdoc` (which looks like it isn't used). Was this intentional?

Ideally, we've found that `@viamrobotics/eslint-config` is easiest to maintain if _all_ its peer dependencies are specified. You can find a list of these in [its README](https://github.com/viamrobotics/js-config/blob/main/packages/eslint-config/README.md):

- @typescript-eslint/parser
- @typescript-eslint/eslint-plugin
- eslint
- eslint-config-prettier
- eslint-plugin-sonarjs
- eslint-plugin-unicorn

Without these peers specified, you run the risk of getting outdated peers even when you're on the latest version of our shared config (we keep the acceptable peer versions wide to minimize downstream disruption to config users)

mcous (approved):
LGTM; looks like we're still got some issues with which eslint plugins are installed via `package.json` but that's a pretty minor thing

package.json
njooma: Ohhh that was a misunderstanding on my part -- I thought this comment was directed at the changed line, not at the tsdoc line 🤦‍♂️ Will fix

PR #224, opened 2023-12-29
RSDK-6037 RemoveLatestMapInfo from SLAM

jeremyrhyde:


maximpertsov (approved):
change looks fine, though we should probably call this out as a breaking change in the next release.

CC @njooma moving forward I wonder if makes sense to add deprecation warnings if we know a method is being removed, or maybe it's fine until we get to 1.0.0

njooma:
> change looks fine, though we should probably call this out as a breaking change in the next release.
> 
> CC @njooma moving forward I wonder if makes sense to add deprecation warnings if we know a method is being removed, or maybe it's fine until we get to 1.0.0

Ya we should have a better process for marking methods as deprecated for a certain amount of time prior to removing. But for now let's post in the channel that this is a breaking change

PR #227, opened 2024-01-04
Update README.md

mestcihazal:
Moved 'npm start' to after setting up env variables otherwise, it doesn't work/recognize and you'll need to restart.

also outside of env variables, you also have to set your component names and make sure it matches  your own robot configuration


mcous (approved):
📜

maximpertsov (approved):

PR #235, opened 2024-01-31
RSDK-6041 - Add dial timeout option and default dial timeout

cheukt:
see https://github.com/viamrobotics/goutils/pull/240

default timeout is 5s

- [x] will need to update rpc version once its out

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "0.9.2",
       "license": "Apache-2.0",
       "dependencies": {
-        "@viamrobotics/rpc": "^0.2.2",
+        "@viamrobotics/rpc": "^0.2.3",
         "exponential-backoff": "^3.1.1"
       },
       "devDependencies": {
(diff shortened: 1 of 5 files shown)


njooma (approved):
LGTM. Only thing is that I would maybe consider creating a constant value for the default timeout, since it's repeated a bunch of times. We could then link the value in the docstring to the constant value.

cheukt:
@njooma added a constants file, not sure if TS has a different convention

njooma:
> @njooma added a constants file, not sure if TS has a different convention

I'm cool with this -- will defer to @maximpertsov since he's the architect behind this.

maximpertsov:
> > @njooma added a constants file, not sure if TS has a different convention
> 
> I'm cool with this -- will defer to @maximpertsov since he's the architect behind this.

Making it a constant is a good idea - I'm indifferent on whether to have a separate file for them or to collocate in dial.ts

maximpertsov (approved):

PR #236, opened 2024-01-31
[RSDK-6497] Vision Service

njooma:
Adds vision service to the TS SDK

### DIFF
--- a/src/services/vision/client.ts
+++ b/src/services/vision/client.ts
@@ -1,0 +1,149 @@
+import { Struct } from 'google-protobuf/google/protobuf/struct_pb';
+import pb from '../../gen/service/vision/v1/vision_pb';
+import { VisionServiceClient } from '../../gen/service/vision/v1/vision_pb_service';
+import type { MimeType } from '../../main';
+import type { RobotClient } from '../../robot';
+import type { Options, StructType } from '../../types';
+import { doCommandFromClient, promisify } from '../../utils';
+import type { Vision } from './vision';
+
+/**
+ * A gRPC-web client for a Vision service.
+ *
+ * @group Clients
+ */
+export class VisionClient implements Vision {
+  private client: VisionServiceClient;
+  private readonly name: string;
+  private readonly options: Options;
+
+  constructor(client: RobotClient, name: string, options: Options = {}) {
+    this.client = client.createServiceClient(VisionServiceClient);
+    this.name = name;
+    this.options = options;
+  }
+
+  private get service() {
+    return this.client;
+  }
+
+  async getDetectionsFromCamera(cameraName: string, extra: StructType = {}) {
+    const { service } = this;
+
+    const request = new pb.GetDetectionsFromCameraRequest();
+    request.setName(this.name);
+    request.setCameraName(cameraName);
+    request.setExtra(Struct.fromJavaScript(extra));
+
+    this.options.requestLogger?.(request);
+
+    const response = await promisify<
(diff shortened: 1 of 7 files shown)


njooma:
@purplenicole730 I am assigning this to you because I will be OOO for the next few days. Assuming this gets approved, could you please squash and merge? Thanks!

src/services/vision/vision.ts
```diff
@@ -0,0 +1,78 @@
...
+   * @param width - The width of the image.
+   * @param height - The height of the image.
+   * @param mimeType - The MimeType of the image.
+   * @returns - The list of Detections.
+   */
+  getDetections: (
+    image: Uint8Array,
+    width: number,
+    height: number,
+    mimeType: MimeType,
```
njooma, line 31: I want to call out that this is different from the way other SDKs do it. Because we're not using a native representation of an image (like `PIL.Image` for python or `img.Image` for golang), we have no way of getting the width/height. So I'm adding them as explicit parameters here.

Cc @bhaney

cheukt (approved):

src/services/vision/vision.ts
njooma: Also, the most important function to get out is `getDetectionsFromCamera`, so if `getDetections` and `getClassifications` aren't 100% because of this issue, it's probably still worth merging this in, and then addressing any of those concerns in a follow-up PR

maximpertsov (approved):

purplenicole730:
I looked it over before squashing and merging.

src/services/vision/vision.ts
bhaney: @njooma @maximpertsov not getting the width or the height is a pretty big deal as most vision service functions will require that in some way or another. getDetectionsFromCamera is usually implemented as a wrapper around getDetections and feeds the images it gets from the camera into getDetections. Are there helper functions that can be added to the SDK for reading the headers of JPEG or PNG images, which usually store the dimensions of the image within them?

bhaney: Technically, all images should be able to be deciphered from their headers (height, width, and mime-type) so it probably is not explicitly necessary to have them as arguments into these functions

src/services/vision/client.ts
```diff
@@ -0,0 +1,149 @@
...
+    extra: StructType = {}
+  ) {
+    const { service } = this;
+
+    const request = new pb.GetClassificationsRequest();
+    request.setName(this.name);
+    request.setImage(image);
+    request.setWidth(width);
+    request.setHeight(height);
+    request.setMimeType(mimeType);
```
bhaney, line 113: I don't think these are necessary - they should be able to be read from the headers as the images, as long as the images are of a known type. It would be better to just have a helper function that could be used to read image headers. This is a common task and there are probably functions already out there

src/services/vision/client.ts
```diff
@@ -0,0 +1,149 @@
...
+    extra: StructType = {}
+  ) {
+    const { service } = this;
+
+    const request = new pb.GetDetectionsRequest();
+    request.setName(this.name);
+    request.setImage(image);
+    request.setWidth(width);
+    request.setHeight(height);
+    request.setMimeType(mimeType);
```
bhaney, line 62: most likely not necessary - these can be inferred from the header of the image bytes

bhaney (commented):
I would prefer height, width, and mime_type not be in the method signatures

bhaney:
@purplenicole730

PR #249, opened 2024-02-27
RSDK-6632 - Typescript SDK throws if MotionConfiguration.obstacleDetectorsList is not specified

nfranczak:
Steps to reproduce :

Typescript SDK encodeMotionConfiguration throws an exception if MotionConfiguration.obstacleDetectorsList is not specified

Expected behavior :

Typescript SDK allows all MotionConfiguration attributes to be optional

### DIFF
--- a/src/services/motion/types.ts
+++ b/src/services/motion/types.ts
@@ -81,13 +81,13 @@ export const encodeMotionConfiguration = (
         obstacleDetector.setCamera(encodeResourceName(od.camera));
       }
       return obstacleDetector;
-    }) ?? new Array<pb.ObstacleDetector>
+    }) ?? new Array<pb.ObstacleDetector>()
   );
-  result.setPositionPollingFrequencyHz(obj.positionPollingFrequencyHz?? 1);
-  result.setObstaclePollingFrequencyHz(obj.obstaclePollingFrequencyHz?? 1);
-  result.setPlanDeviationM(obj.planDeviationM?? 2.6);
-  result.setLinearMPerSec(obj.linearMPerSec?? 0.3);
-  result.setAngularDegsPerSec(obj.angularDegsPerSec?? 20);
+  result.setPositionPollingFrequencyHz(obj.positionPollingFrequencyHz ?? 1);
+  result.setObstaclePollingFrequencyHz(obj.obstaclePollingFrequencyHz ?? 1);
+  result.setPlanDeviationM(obj.planDeviationM ?? 2.6);
+  result.setLinearMPerSec(obj.linearMPerSec ?? 0.3);
+  result.setAngularDegsPerSec(obj.angularDegsPerSec ?? 20);
 
   return result;
 };
--- a/src/services/motion/types.ts
+++ b/src/services/motion/types.ts
@@ -6,7 +6,7 @@ export type ObstacleDetector = pb.ObstacleDetector.AsObject;
 export type LinearConstraint = pb.LinearConstraint.AsObject;
 export type OrientationConstraint = pb.OrientationConstraint.AsObject;
 export type CollisionSpecification = pb.CollisionSpecification.AsObject;
-export type MotionConfiguration = pb.MotionConfiguration.AsObject;
+export type MotionConfiguration = Partial<pb.MotionConfiguration.AsObject>;
 export type GetPlanResponse = pb.GetPlanResponse.AsObject;
 export type ListPlanStatusesResponse = pb.ListPlanStatusesResponse.AsObject;
 type ValueOf<T> = T[keyof T];
@@ -72,7 +72,7 @@ export const encodeMotionConfiguration = (
   const result = new pb.MotionConfiguration();
 
   result.setObstacleDetectorsList(
-    obj.obstacleDetectorsList.map((od: ObstacleDetector) => {
+    obj.obstacleDetectorsList?.map((od: ObstacleDetector) => {
       const obstacleDetector = new pb.ObstacleDetector();
       if (od.visionService) {
         obstacleDetector.setVisionService(encodeResourceName(od.visionService));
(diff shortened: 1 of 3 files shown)


nfranczak:
relates to https://github.com/viamrobotics/rdk/pull/3574

src/services/motion/types.ts
```diff
@@ -72,7 +72,7 @@ export const encodeMotionConfiguration = (
   const result = new pb.MotionConfiguration();
 
   result.setObstacleDetectorsList(
-    obj.obstacleDetectorsList.map((od: ObstacleDetector) => {
+    obj.obstacleDetectorsList?.map((od: ObstacleDetector) => {
```
micheal-parks, line 75: This is a bit hard to read . Instead I'd recommend:
```ts
const { obstacleDetectorsList = [] } = obj;
result.setObstacleDetectorsList(obstacleDetectorsList.map(...))
```
We can then remove the `new Array` coalescing below on line 84.

src/services/motion/types.ts
```diff
@@ -81,13 +81,13 @@ export const encodeMotionConfiguration = (
...
       return obstacleDetector;
-    })
+    }) ?? new Array<pb.ObstacleDetector>()
   );
-  result.setPositionPollingFrequencyHz(obj.positionPollingFrequencyHz);
-  result.setObstaclePollingFrequencyHz(obj.obstaclePollingFrequencyHz);
-  result.setPlanDeviationM(obj.planDeviationM);
-  result.setLinearMPerSec(obj.linearMPerSec);
-  result.setAngularDegsPerSec(obj.angularDegsPerSec);
+  result.setPositionPollingFrequencyHz(obj.positionPollingFrequencyHz ?? 1);
```
micheal-parks, line 86: Are these necessary to coalesce? In other words, will this break somewhere if we do not send default values? My concern is that default values now potentially exist in multiple places (SDK and robot).

nfranczak: >Are these necessary to coalesce? In other words, will this break somewhere if we do not send default values?... My concern is that default values now potentially exist - and would need to be maintained - in multiple places (SDK and robot).

Good catch! I will change this to be:
```
  if (obj.angularDegsPerSec) {
    result.setAngularDegsPerSec(obj.angularDegsPerSec);
  }
```

src/services/motion/types.ts
```diff
@@ -83,11 +84,21 @@ export const encodeMotionConfiguration = (
       return obstacleDetector;
     })
   );
-  result.setPositionPollingFrequencyHz(obj.positionPollingFrequencyHz);
-  result.setObstaclePollingFrequencyHz(obj.obstaclePollingFrequencyHz);
-  result.setPlanDeviationM(obj.planDeviationM);
-  result.setLinearMPerSec(obj.linearMPerSec);
-  result.setAngularDegsPerSec(obj.angularDegsPerSec);
+  if (obj.positionPollingFrequencyHz) {
```
micheal-parks, line 87: It would be safer for these checks to do:

```ts
if ('positionPollingFrequencyHz' in obj) {
  result.setPositionPollingFrequencyHz(obj.positionPollingFrequencyHz)
}
```
The current check will evaluate to false if any of these props are zero, and therefore won't be set, which will cause unexpected behavior.

purplenicole730 (approved):

src/services/motion/types.ts
```diff
@@ -71,8 +71,9 @@ export const encodeMotionConfiguration = (
 ): pb.MotionConfiguration => {
   const result = new pb.MotionConfiguration();
 
+  const { obstacleDetectorsList = [] } = obj;
```
nicksanford, line 74: How does this behave:
1.  if `obj` is `undefined`?
2. if `obstacleDetectorsList` is not a member of `obj`?

nicksanford:
@nfranczak Tests are not passing

src/services/motion/types.ts
micheal-parks: 1. The parameter type for `obj` for this function states that it will `never` be undefined. If that's not the case, it should be an optional param, and the param should recieve a default value:
```ts
export const encodeMotionConfiguration = (
  obj?: MotionConfiguration = {}
): pb.MotionConfiguration =>
```
2. If it's not a member of this object, it will be assigned a default value of `[]`.

src/services/motion/types.ts
nfranczak: with this, when linting, I run into:
```
src/services/motion/types.ts(88,42): error TS2345: Argument of type 'number | undefined' is not assignable to parameter of type 'number'.
  Type 'undefined' is not assignable to type 'number'.
```
I will change the conditionals to be:
```
if (obj.positionPollingFrequencyHz !== undefined) {
    result.setPositionPollingFrequencyHz(obj.positionPollingFrequencyHz);
}
```
this way even if any of the props are zero or false they are set.

micheal-parks: That works!

src/services/motion/types.ts
```diff
@@ -67,12 +67,13 @@ export const encodeConstraints = (obj: Constraints): pb.Constraints => {
 };
 
 export const encodeMotionConfiguration = (
-  obj: MotionConfiguration
+  obj: MotionConfiguration = {}
```
nfranczak, line 70: @micheal-parks since I added this, do you think I should remove 
```
    if (motionConfig) {
      request.setMotionConfiguration(encodeMotionConfiguration(motionConfig));
    }
```
from `src/services/motion/client.ts`?

nicksanford (approved):
Please confirm this change works with the RDK RC card post merge.

src/services/motion/types.ts
micheal-parks: I think it still makes sense to not call this function if no config is provided

micheal-parks:
Pre-merge might be difficult with our current setup, I think post-merge is fine for now. All you'll have to do is:
1. bump the SDK version in this repo's package.json. That'll trigger our CI pipeline to ship a new version to NPM
2. install the new version in the RDK, use this method, and run npm run check. Nick S might be able to assist since he already started work here.

micheal-parks (approved):
Thanks for addressing my comments, good stuff!

PR #255, opened 2024-03-06
Initial ideas for TS instrumentation (`getStats`)

benjirewis:
Adds a `getStats` method to `RobotClient`.


benjirewis:
Closing for now; after discussing offline with @cheukt and @mattjperez , I don't think we'll want to use a new `getStats` method at all for initial metric gathering.

PR #266, opened 2024-03-25
RSDK-6838: add provisioning wrappers

purplenicole730:
added provisioning wrappers and unit tests

*has not been tested end-to-end as it was concluded unit tests would be enough*


stuqdog (approved):
lgtm!

njooma (approved):

PR #272, opened 2024-04-10
RSDK-7200: add ml training wrapper

purplenicole730:
Added the ML Training wrappers and tests

After some discussion, also added `provisioning` files to the app folder.


src/app/ml-training-client.ts
```diff
@@ -0,0 +1,96 @@
...
+export class MlTrainingClient {
+  private service: MLTrainingServiceClient;
+
+  constructor(serviceHost: string, grpcOptions: RpcOptions) {
+    this.service = new MLTrainingServiceClient(serviceHost, grpcOptions);
+  }
+
+  async submitTrainingJob(
+    datasetId: string,
+    orgId: string,
```
njooma, line 21: Can we make `orgId` the first parameter? I feel like organization should always be first, since that's the main way we separate information

src/app/provisioning-client.ts
```diff
@@ -1,28 +1,33 @@
...
 
+  /**
+   * Get the status of the Smart Machine.
+   *
+   * @returns The Smart Machine status
+   */
   async getSmartMachineStatus() {
     const { service } = this;
     const request = new pb.GetSmartMachineStatusRequest();
-    this.options.requestLogger?.(request);
```
njooma, line 25: Why remove the loggers?

purplenicole730: More in line with the other app clients (originally I originally based it on a service instead)

njooma (approved):

PR #276, opened 2024-04-16
RSDK-7215: add billing wrapper

purplenicole730:


src/app/billing-client.ts
```diff
@@ -0,0 +1,97 @@
...
+      ? new Date(
+          startDate.getSeconds() * 1000 + startDate.getNanos() / 1_000_000
+        )
+      : undefined,
+    end: endDate
+      ? new Date(endDate.getSeconds() * 1000 + endDate.getNanos() / 1_000_000)
+      : undefined,
+  };
+  return result;
+};
```
njooma, line 36: Do you really need all of this? I think `proto.getStartDate()` has a `toDate()` function that you can use instead.

src/app/billing-client.ts
```diff
@@ -0,0 +1,97 @@
...
+
+    const req = new pb.GetInvoicePdfRequest();
+    req.setId(id);
+    req.setOrgId(orgId);
+
+    const response = await promisify<
+      pb.GetInvoicePdfRequest,
+      pb.GetInvoicePdfResponse
+    >(service.getInvoicePdf.bind(service), req);
+    return response.getChunk();
```
njooma, line 95: I believe this is a stream of chunks. So you will have to listen to the whole stream and aggregate these chunks into a single object before returning

njooma (changes requested):

stuqdog (commented):
Got a question about the `decodeMonthUsage` method, otherwise lgtm

src/app/billing-client.ts
purplenicole730: Thank you for reminding me! I was using it before, but it was failing for something else and so I did it manually. I forgot to change it back.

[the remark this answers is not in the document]
purplenicole730: I was running into some problems with testing, so did everything manually. Your suggestion actually got me to realize what was wrong! Thank you, I guess I just needed to look at it again with fresh eyes.

src/app/billing-client.ts
purplenicole730: Thank you!

src/app/billing-client.ts
```diff
@@ -0,0 +1,117 @@
+import { type RpcOptions } from '@improbable-eng/grpc-web/dist/typings/client.d';
+import { BillingServiceClient } from '../gen/app/v1/billing_pb_service';
+import pb from '../gen/app/v1/billing_pb';
+import { promisify } from '../utils';
+
+type GetCurrentMonthUsageResponse =
+  Partial<pb.GetCurrentMonthUsageResponse.AsObject> & {
+    start?: Date;
+    end?: Date;
+  };
```
stuqdog, line 10: (q) pardon my typescript ignorance, why is this necessary? Can we not set `start` or `end` on a `GetCurrentMonthUsageResponse` object without doing this?

stuqdog (approved):
One question for my own knowledge, but LGTM!

njooma (approved):

src/app/billing-client.ts
purplenicole730: Good question! We have to define the properties in Typescript. Imo, it also helps users to figure out what the actual property is instead of adding it without any mention outside of the function itself.

PR #277, opened 2024-04-17
RSDK-7230: add machine_id to cloud metadata

maximpertsov:
~Breaking change!~ Not anymore!

Add `machineId` to the `getCloudMetadata` robot client method. ~Also change the return type and rename `robotPartId` to `machinePartId` for clarity.~ [We are just deprecating the duplicate field instead](https://github.com/viamrobotics/viam-python-sdk/pull/585#discussion_r1571261813).

Depends on:
* [x] https://github.com/viamrobotics/api/pull/487

### DIFF
--- a/src/services/motion/client.test.ts
+++ b/src/services/motion/client.test.ts
@@ -43,14 +43,12 @@ describe('moveOnGlobe', () => {
       type: 'component',
       subtype: 'base',
       name: 'myBase',
-      machinePartId: '',
     };
     const expectedMovementSensorName = {
       namespace: 'viam',
       type: 'component',
       subtype: 'movementsensor',
       name: 'myMovementsensor',
-      machinePartId: '',
     };
     const expectedMotionConfiguration = undefined;
     const expectedExtra = {};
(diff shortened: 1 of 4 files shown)


cheukt (approved):

purplenicole730 (approved):
LGTM. I'm guessing the parsing out instead of using `AsObject` is to avoid returning the redundant `robotPartId`?

maximpertsov:
> LGTM. I'm guessing the parsing out instead of using `AsObject` is to avoid returning the redundant `robotPartId`?

yeah... I might do this differently depending on the answer to [my question](https://github.com/viamrobotics/viam-python-sdk/pull/585#discussion_r1569267606) on the corresponding python sdk change.

src/services/motion/client.test.ts
```diff
@@ -43,7 +43,6 @@ describe('moveOnGlobe', () => {
       type: 'component',
       subtype: 'base',
       name: 'myBase',
-      machinePartId: '',
```
maximpertsov, line 46: removing a reference to [a field that was removed in an unrelated proto update](https://github.com/viamrobotics/api/commit/274901dc0c4735cadb75792617c0700e8cde7bda)

cheukt (approved):

PR #283, opened 2024-05-01
RSDK-6526 - Add app client wrappers

stuqdog:
Adds app wrappers and tests to typescript SDK

Note that `uploadModuleFile` remains unimplemented here. [Client streaming is not supported by default](https://github.com/improbable-eng/grpc-web/blob/master/client/grpc-web/docs/websocket.md) in http/2-based transport in browsers. There _is_ a `websocket`-based solution but it would require updating on both the server side and client side, and according to at least [one source](https://jbrandhorst.com/post/state-of-grpcweb/#f19) is not recommended for production use (admittedly this seems to be from 4 years ago, but I haven't found evidence to the contrary yet). All of this led me to believe that this is a complicated enough project to justify separating into a separate PR (assuming it's worth doing at all).

Happy to discuss/revisit if there's any disagreement!


njooma (approved):
Agree, the client streaming is too big a project to include with right now. LGTM!

purplenicole730 (approved):
Thanks for doing the research needed to get this done! Now at least we can make an informed opinion and path forward

src/app/viam-client.ts
```diff
@@ -34,6 +35,7 @@ export class ViamClient {
   private serviceHost: string;
 
   public dataClient: DataClient | undefined;
+  public appClient: AppClient | undefined;
```
purplenicole730, line 38: The `viam-client.test.ts` file also has tests to show that the ViamClient has these created when the `createViamClient()` function is called. 
`expect(client.appClient).toBeInstanceOf(AppClient);`

PR #285, opened 2024-05-03
RSDK-7199: add data sync wrappers

purplenicole730:
NOTE: `fileUpload` and `streamingDataCaptureUpload` are not included, as we decided that client streaming in Typescript is out of scope of these wrapper tickets right now.


src/app/data-client.ts
```diff
@@ -709,6 +714,147 @@ export class DataClient {
...
+    const sensorContents: dataSyncPb.SensorData[] = [];
+    for (const [i, data] of tabularData.entries()) {
+      const sensorData = new dataSyncPb.SensorData();
+
+      const sensorMetadata = new dataSyncPb.SensorMetadata();
+      const dates = dataRequestTimes[i];
+      if (dates) {
+        sensorMetadata.setTimeRequested(Timestamp.fromDate(dates[0]));
+        sensorMetadata.setTimeReceived(Timestamp.fromDate(dates[1]));
+      }
```
purplenicole730, line 773: I have spent an insane amount of time trying to make lint happy. Things I have tried:
- `if (dates)` "'Date | undefined' is not assignable to parameter of type 'Date'"
- `if (dates && dates[0] && dates[1])` "Prefer using an optional chain expression instead, as it's more concise and easier to read"
- `Timestamp.fromDate(dates?.[0])` "Unnecessary optional chain on a non-nullish value"
-

src/app/data-client.ts
```diff
@@ -709,6 +714,157 @@ export class DataClient {
...
+   *   as an array.
+   * @param partId The part ID of the component used to capture the data
+   * @param componentType The type of the component used to capture the data
+   *   (e.g., "movementSensor")
+   * @param componentName The name of the component used to capture the data
+   * @param methodName The name of the method used to capture the data.
+   * @param tags The list of tags to allow for tag-based filtering when
+   *   retrieving data
+   * @param dataRequestTimes Array of Date arrays, each containing two `Date`
+   *   objects denoting the times this data was requested[0] by the robot and
```
stuqdog, line 735: instead of an array of arrays, can we have this be an array of `[Datetime, Datetime]` tuples, so we don't invite the possibility of people including an incorrect amount of data?

src/app/data-client.ts
```diff
@@ -709,6 +714,157 @@ export class DataClient {
...
+    metadata.setTagsList(tags ?? []);
+
+    const sensorContents: dataSyncPb.SensorData[] = [];
+    for (const [i, data] of tabularData.entries()) {
+      const sensorData = new dataSyncPb.SensorData();
+
+      const sensorMetadata = new dataSyncPb.SensorMetadata();
+      const dates = dataRequestTimes[i];
+      if (dates) {
+        if (dates[0]) {
```
stuqdog, line 771: The comment makes it seem that each entry in `dataRequestTimes` must have two `Datetime` objects but here we're checking that they exist. Is that check necessary? If so, should we be updating the type definition/comment to clarify that?

src/app/data-client.ts
```diff
@@ -709,6 +714,157 @@ export class DataClient {
...
+   */
+  async binaryDataCaptureUpload(
+    binaryData: Uint8Array,
+    partId: string,
+    componentType: string,
+    componentName: string,
+    methodName: string,
+    fileExtension: string,
+    tags?: string[],
+    dataRequestTimes?: Date[]
```
stuqdog, line 827: Similar comments as above re: making this a tuple and whether or not we need to check below for the existence of particular entries.

stuqdog (commented):
Nice, this is looking good! I have some questions/suggestions around the typing of the `dataRequestTimes` parameters but otherwise this seems great.

src/app/data-client.ts
purplenicole730: I hate that I had to do this but I had to for lint reasons.

stuqdog: Ahh, that makes sense. I'm not positive but I suspect that would no longer be true if the type was a tuple rather than an array.

njooma (approved):
LGTM % @stuqdog's changes. I agree with his comments

src/app/data-client.ts
purplenicole730: Ah, yes. I remember why I changed this from `[Date, Date][]` to an array of arrays. I was getting an error saying: `src/app/data-client.test.ts(1007,10): error TS2322: Type 'Date[]' is not assignable to type '[Date, Date]'.`.

I agree that having a tuple is the way to go. It does means that when someone is trying to fill in this value, they can't just do:
```
const var = [date1, date2]
await client.tabularDataCaptureUpload(..., var)
```
They now have to declare the type: `const var: [Date, Date] = [date1, date2]`. I think this is okay, because while it is more annoying, it does make it more purposeful.

purplenicole730: Oh, that was a listing error, so users can still just use the simple `const var = [date1, date2]`!

stuqdog (approved):
Nice, lgtm!

PR #286, opened 2024-05-07
Export disconnect and reconnect events

ethanlook:
See https://github.com/viamrobotics/app/pull/4672. I need to listen to the disconnect and reconnect events - I could hardcode the strings, but it would be more convenient if I could just import the constants.

I'm not tied to this, so please let me know if this isn't the right way to go about it!

### DIFF
--- a/examples/vanilla/src/main.ts
+++ b/examples/vanilla/src/main.ts
@@ -57,13 +57,13 @@ async function run(client: VIAM.RobotClient) {
 
 // This function is called when the robot is disconnected.
 // Feel free to replace it with whatever logic you want to test out!
-async function disconnected(event) {
+async function disconnected() {
   console.log('The robot has been disconnected. Trying reconnect...');
 }
 
 // This function is called when the robot is reconnected.
 // Feel free to replace it with whatever logic you want to test out!
-async function reconnected(event) {
+async function reconnected() {
   console.log('The robot has been reconnected. Work can be continued.');
 }
 
@@ -73,8 +73,8 @@ async function main() {
   try {
     client = await connect();
     console.log('connected!');
-    client.on('disconnected', disconnected);
-    client.on('reconnected', reconnected);
+    client.on(VIAM.MachineConnectionEvent.DISCONNECTED, disconnected);
+    client.on(VIAM.MachineConnectionEvent.RECONNECTED, reconnected);
   } catch (error) {
     console.log(error);
     return;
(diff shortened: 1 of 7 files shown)


njooma (approved):
I think this should be fine. I was thinking about if we should export them named as something else like `MACHINE_STATUS.RECONNECTED` or something?

src/main.ts
```diff
@@ -475,3 +475,8 @@ export type {
 export * from './types';
 
 export { doCommandFromClient, promisify } from './utils';
+
+export {
+  RECONNECTED,
+  DISCONNECTED,
+} from './events';
```
ethanlook, line 482: (commenting here for threading)

@njooma it would be a bigger change, but we could define it as an enum. I think that would be the most clear for usage and naming for export. We could consider adding `connecting`, `connected`, etc here as well.

```
export enum MachineConnectionEvent {
  RECONNECTED = 'reconnected',
  DISCONNECTED = 'disconnected',
}
```

stuqdog (approved):
This looks reasonable to me, don't think I have sufficient context to weigh in on whether the benefits of an enum outweigh the cost.

src/main.ts
njooma: Ya agreed, I think we should add it now. Also would it really be a big change? Seems like we just add your `enum` to `events.ts` and export that, right? We don't actually have to change anything else since the string values are still the same

ethanlook: I would want to start using the enum internally as well - not crazy but a bigger diff.

njooma: Ya for sure, but also the SDK team could take care of that if needed. Probably is easier/faster for you to do it now, but seemed out of scope

examples/vanilla/src/main.ts
```diff
@@ -57,13 +57,13 @@ async function run(client: VIAM.RobotClient) {
 
 // This function is called when the robot is disconnected.
 // Feel free to replace it with whatever logic you want to test out!
-async function disconnected(event) {
```
ethanlook, line 60: Was getting warnings that these events are unused. Happy to add them back.

src/main.ts
ethanlook: Quickly made this change, let me know what you think.

ethanlook:
I want to take a second to recognize the `Ethan-to-Ethan` code review here. Thank you, @stuqdog

stuqdog (approved):

njooma (approved):

PR #293, opened 2024-05-10
Enable workflows for all pull requests

ethanlook:
Let me know what you think about this change - we do the same thing in App as well. This makes it so we can stack PRs but still have workflows run on all PRs in the stack. I'm flexible, let me know if this change doesn't make sense!


njooma (approved):
LGTM

PR #292, opened 2024-05-10
RSDK-7245: reconnect with backoff on initial connection

ethanlook:
This re-attempts initial connection when calling `createRobotClient` according to backoff parameters. If `noReconnect` is specified it does not re-attempt.

1. https://github.com/viamrobotics/viam-typescript-sdk/pull/291
2. https://github.com/viamrobotics/viam-typescript-sdk/pull/292
3. https://github.com/viamrobotics/viam-typescript-sdk/pull/294

## Change log

- Reduce cognitive complexity of `createRobotClient`
- Use `exponential-backoff` to attempt reconnects if not explicitly disabled with `noReconnect`
- Clean up reconnect `exponential-backoff` usage

## Testing

Not sure what the testing practices here are - let me know what would be best!

PR #291, opened 2024-05-10
RSDK-7244: Emit disconnect events on data channel close

ethanlook:
This fixes RSDK-7244, which describes how the TS SDK does not attempt reconnect when a disconnect occurs.

The TS SDK attempts to reconnect to the peer in a [backoff loop](https://github.com/viamrobotics/viam-typescript-sdk/blob/f49f6857de9277f72f1e4894b985ea976ae51ea0/src/robot/client.ts#L186). This happens whenever a [disconnect event](https://github.com/viamrobotics/viam-typescript-sdk/blob/f49f6857de9277f72f1e4894b985ea976ae51ea0/src/robot/client.ts#L177) occurs.

The TS SDK currently listens to `iceconnectionstatechange` `close` events on the `RTCPeerConnection` to [emit the disconnect](https://github.com/viamrobotics/viam-typescript-sdk/blob/f49f6857de9277f72f1e4894b985ea976ae51ea0/src/robot/client.ts#L469-L483) necessary to start the reconnection. These events are never being emitted, as confirmed with browser WebRTC logging. `connectionstatechange` events are not emitted either.

Instead, we must listen to the `RTCDataChannel` `close` events which is actually what goutils [already does](https://github.com/viamrobotics/goutils/blob/612b3f81b6b71c94b89805e63390b6ec5336d8ca/rpc/js/src/ClientChannel.ts#L38) to know when to close the `RTCPeerConnection`.

https://github.com/viamrobotics/goutils/pull/287: exposes the `RTCDataChannel`
https://github.com/viamrobotics/viam-typescript-sdk/pull/291: **this PR** listens to `close` events on the `RTCDataChannel`

### TODO

- [x] Bump to `@viamrobotics/rpc@0.2.5` once https://github.com/viamrobotics/goutils/pull/287 is merged and published

### PR stack

Additional changes are stacked on top of this PR but are not necessary for the disconnect fix.

1. https://github.com/viamrobotics/viam-typescript-sdk/pull/291
2. https://github.com/viamrobotics/viam-typescript-sdk/pull/292
3. https://github.com/viamrobotics/viam-typescript-sdk/pull/294

PR #294, opened 2024-05-13
RSDK-7413: Remove global event dispatcher

ethanlook:
This removes the global `EventDispatcher` instance and makes each `RobotClient` responsible for its own event dispatching.

1. https://github.com/viamrobotics/viam-typescript-sdk/pull/291
2. https://github.com/viamrobotics/viam-typescript-sdk/pull/292
3. https://github.com/viamrobotics/viam-typescript-sdk/pull/294

## Change log

- Remove the global `EventDispatcher`
- Clean up instances of calling `events.emit` and `events.on`
- Pass `onDisconnect` handler to `GrpcConnectionManager`
- Fix up tests

## Testing

- Unit tests
- `examples/connect-app` locally

### DIFF
--- a/src/events.ts
+++ b/src/events.ts
@@ -42,5 +42,3 @@ export class EventDispatcher {
     }
   }
 }
-
-export const events = new EventDispatcher();
--- a/src/extra/stream/client.test.ts
+++ b/src/extra/stream/client.test.ts
@@ -4,24 +4,25 @@ import { vi, beforeEach, afterEach, describe, expect, it } from 'vitest';
 import { RobotClient } from '../../robot';
 vi.mock('../../robot');
 
-import { events } from '../../events';
 import { StreamServiceClient } from '../../gen/proto/stream/v1/stream_pb_service';
 vi.mock('../../gen/proto/stream/v1/stream_pb_service');
 
 import { StreamClient } from './client';
-
-let streamClient: StreamClient;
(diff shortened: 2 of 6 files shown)


njooma (commented):
Q: In TS/JS, can any object `emit` and receive events using `on`? For example, your [updates in stream client](https://github.com/viamrobotics/viam-typescript-sdk/pull/294/files#diff-d9ba4160980d47b6e35dc3227b64adf7fdddb5e4615f7a42768689481aa41bb3R38) have the `StreamServiceClient` receiving events using `on`, but I don't see where the client actually sends those events?

ethanlook:
> Q: In TS/JS, can any object `emit` and receive events using `on`? For example, your [updates in stream client](https://github.com/viamrobotics/viam-typescript-sdk/pull/294/files#diff-d9ba4160980d47b6e35dc3227b64adf7fdddb5e4615f7a42768689481aa41bb3R38) have the `StreamServiceClient` receiving events using `on`, but I don't see where the client actually sends those events?

The [`RobotClient` extends `EventDispatcher`](https://github.com/viamrobotics/viam-typescript-sdk/blob/main/src/robot/client.ts#L86) which gives it the `emit`/`on` behavior! It already extended that, so this is just flowing all events through the `RobotClient` instead of the global instance.

njooma (approved):

purplenicole730 (approved):

PR #298, opened 2024-05-16
RSDK-7153 ReadAnalogReaderResponse update

oliviamiller:


purplenicole730 (approved):

src/components/board/board.ts
```diff
@@ -11,6 +11,12 @@ export interface Tick {
   time: number;
 }
 export type Duration = PBDuration.AsObject;
+export interface AnalogValue {
```
njooma, line 14: can we just make this `export type AnalogValue = pb.ReadAnalogReaderResponse.AsObject`? I'm not 100% sure it'll work but I would like to typealias it since it doesn't introduce any new values

njooma (changes requested):

src/components/board/board.ts
oliviamiller: it works, changed

src/components/board/client.ts
```diff
@@ -138,7 +138,13 @@ export class BoardClient implements Board {
...
       pb.ReadAnalogReaderResponse
     >(boardService.readAnalogReader.bind(boardService), request);
-    return response.getValue();
+    const value: AnalogValue = {
+      value: response.getValue(),
+      minRange: response.getMinRange(),
+      maxRange: response.getMaxRange(),
+      stepSize: response.getStepSize(),
+    };
+    return value;
```
njooma, line 147: can we not just return `response.asObject()` here? (or whatever converts a `pb.ReadAnalogReaderResponse` into a JS Object?)

oliviamiller: I dont think so, just tried and got `Unsafe return of an `any` typed value.` error

purplenicole730: Did you also try `return response.toObject()`?

oliviamiller: it says thats not a function

purplenicole730: That's... weird. Sorry, I've been spending the last 10 minutes trying to figure out why it would say it's not a function... 

I tried it on my end. (Forgot I had to `make all` first to get the latest proto changes). I also switched the `board.ts` to return a `Promise<AnalogValue>;`.
and changed this to: `return response.toObject();`. It worked for me

purplenicole730: image
image
I don't know if this helps, but these are my changes. I ran `make format lint` without a problem.

oliviamiller: Did you run the test? the test fails with the error `AssertionError: promise rejected "TypeError: response.toObject is not a function instead of resolving`

purplenicole730: Ohhhh yes, that makes sense! Because in your test, you emulate what the other `get<>` functions are, but NOT `toObject()`. So you would have to change the mock to be like something similar to this:
```
  BoardServiceClient.prototype.readAnalogReader = vi
    .fn()
    .mockImplementation((_req, _md, cb) => {
      cb(null, { toObject: () => testAnalogValue });
    });
```

purplenicole730: The thing you're mocking is the actual **pb.response** type itself. So you have to manually define what the functions of your mock response type would return. Now that the function `readAnalogReader()` is returning `response.toObject()` instead of all the other get functions, that's what you have to mock instead.
Let me know if you don't get it! It took me a while to wrap my head around it too.

oliviamiller: ohhhh okay i see, that worked thanks so much!

src/components/board/client.test.ts
```diff
@@ -13,6 +13,47 @@ import { BoardServiceClient } from '../../gen/component/board/v1/board_pb_servic
...
+    .mockImplementation(() => new BoardServiceClient('mysensor'));
+
+  BoardServiceClient.prototype.readAnalogReader = vi
+    .fn()
+    .mockImplementation((_req, _md, cb) => {
+      cb(null, {
+        getValue: () => testValue,
+        getMinRange: () => testAnalogMin,
+        getMaxRange: () => testAnalogMax,
+        getStepSize: () => testStepSize,
```
purplenicole730, line 40: I suspect you no longer need to define these since the client function isn't calling them anymore

src/components/board/board.ts
```diff
@@ -11,6 +11,7 @@ export interface Tick {
   time: number;
 }
 export type Duration = PBDuration.AsObject;
+export type AnalogValue = pb.ReadAnalogReaderResponse.AsObject;
```
purplenicole730, line 14: I'm assuming since you defined this because users will need to be able to access this. I forgot to mention, please add this type to the `main.ts` file.

purplenicole730 (changes requested):

src/components/board/board.ts
oliviamiller: is ` tick` supposed to be in that file too? I did that change a couple weeks and didnt add it

purplenicole730: If `tick` and `AnalogValue` are exported types that users have to have access to, then yes.

oliviamiller: okay will add them both, thanks

purplenicole730 (approved):

njooma (approved):

PR #300, opened 2024-05-17
Update vanilla example to use Vite + TS app for robot client

ethanlook:
I've been using this simple Vite + TypeScript app to test during local development and figured it's probably more broadly useful. Please let me know if you don't like this in any way, I'm happy to make changes, abandon it, etc.

teleop-vanilla

1. https://github.com/viamrobotics/viam-typescript-sdk/pull/299
2. ➡️ https://github.com/viamrobotics/viam-typescript-sdk/pull/300

## Change log

- `npm create vite@latest teleop-vanilla`
- Brief README
- Limited connect / disconnect / log resources UI
- Edit: moved to `vanilla`, no longer `teleop-vanilla`


njooma (commented):
I like this, but I think maybe we just replace the existing "vanilla" example with this one? This seems more fully featured and actually how users would use our SDK

njooma (approved):

PR #316, opened 2024-06-24
RSDK-7987 - Arm should re-export JointPositions type from proto

zaporter-work:
https://viam.atlassian.net/browse/RSDK-7987

This follows the pattern from Encoder which also wraps the encoder `pb.PositionType` with a new `EncoderPositionType`

### DIFF
--- a/src/components/arm/arm.ts
+++ b/src/components/arm/arm.ts
@@ -2,7 +2,7 @@ import type { JointPositions } from '../../gen/component/arm/v1/arm_pb';
 
 import type { Pose, Resource, StructType } from '../../types';
 
-export type ArmJointPositions = JointPositions;
+export type ArmJointPositions = JointPositions.AsObject;
 
 /** Represents a physical robot arm that exists in three-dimensional space. */
 export interface Arm extends Resource {
--- a/src/components/arm/arm.ts
+++ b/src/components/arm/arm.ts
@@ -2,6 +2,8 @@ import type { JointPositions } from '../../gen/component/arm/v1/arm_pb';
 
 import type { Pose, Resource, StructType } from '../../types';
 
+export type ArmJointPositions = JointPositions;
+
 /** Represents a physical robot arm that exists in three-dimensional space. */
 export interface Arm extends Resource {
   /** Get the position of the end of the arm expressed as a pose */
@@ -26,7 +28,7 @@ export interface Arm extends Resource {
   ) => Promise<void>;
 
   /** Gets the current position of each joint. */
-  getJointPositions: (extra?: StructType) => Promise<JointPositions>;
+  getJointPositions: (extra?: StructType) => Promise<ArmJointPositions>;
 
   /** Stops the motion of the arm. */
   stop: (extra?: StructType) => Promise<void>;
--- a/src/components/arm.ts
+++ b/src/components/arm.ts
@@ -1,2 +1,2 @@
-export type { Arm } from './arm/arm';
+export type { Arm, ArmJointPositions } from './arm/arm';
 export { ArmClient } from './arm/client';
--- a/src/components/arm/client.ts
+++ b/src/components/arm/client.ts
@@ -99,7 +99,7 @@ export class ArmClient implements Arm {
     if (!result) {
       throw new Error('no pose');
     }
-    return result;
+    return result.toObject();
   }
 
   async stop(extra = {}) {
--- a/src/main.ts
+++ b/src/main.ts
@@ -120,7 +120,7 @@ export { type BillingClient } from './app/billing-client';
  * @group Raw Protobufs
  */
 export { default as armApi } from './gen/component/arm/v1/arm_pb';
-export { type Arm, ArmClient } from './components/arm';
+export { type Arm, type ArmJointPositions, ArmClient } from './components/arm';
 
 /**
  * Raw Protobuf interfaces for a Base component.


src/components/arm/arm.ts
```diff
@@ -2,6 +2,8 @@ import type { JointPositions } from '../../gen/component/arm/v1/arm_pb';
 
 import type { Pose, Resource, StructType } from '../../types';
 
+export type ArmJointPositions = JointPositions;
```
ethanlook, line 5: Can you not use JointPositions from the exported `armApi`?

https://github.com/viamrobotics/viam-typescript-sdk/blob/main/src/main.ts#L122

zaporter-work: That has a:
```
 @deprecated Use {@link ArmClient} instead.
 * @alpha
 * @group Raw Protobufs
```
comment. 

Should I remove that comment?

ethanlook: Hah my bad. I'm not sure, I defer to the opinion of the SDK folks.

njooma: Could we not just re-export `JointPositions`? Does it have to be renamed here? (pardon my ignorance...)

zaporter-work: @njooma 

Not sure -- the SDK is slightly inconsistent as to whether we rename before exporting. But I think it is mostly based on whether we expect the name to be re-used in another component. Because `JointPositions` is defined in the arm.proto, it seemed right to rename it with a prefix... ?

See:
https://github.com/viamrobotics/viam-typescript-sdk/blob/83a1bd0791d72618aab44bc98aedb819552dbd0c/src/components/movement-sensor/movement-sensor.ts
```
export type MovementSensorPosition = pb.GetPositionResponse.AsObject;
export type MovementSensorProperties = pb.GetPropertiesResponse.AsObject;
export type MovementSensorAccuracy = pb.GetAccuracyResponse.AsObject;
```
https://github.com/viamrobotics/viam-typescript-sdk/blob/83a1bd0791d72618aab44bc98aedb819552dbd0c/src/components/encoder/encoder.ts
```
export type EncoderProperties = pb.GetPropertiesResponse.AsObject;

type ValueOf<T> = T[keyof T];
export const EncoderPositionType = pb.PositionType;
export type EncoderPositionType = ValueOf<typeof pb.PositionType>;
```
But:
https://github.com/viamrobotics/viam-typescript-sdk/blob/83a1bd0791d72618aab44bc98aedb819552dbd0c/src/components/board/board.ts
```
export type Duration = PBDuration.AsObject;
export type AnalogValue = pb.ReadAnalogReaderResponse.AsObject;
```

zaporter-work: New to this SDK and haven't done SDK work in a while -- so if you think it is best to not rename, happy to remove that.

maximpertsov: RE: https://github.com/viamrobotics/viam-typescript-sdk/pull/316/files#r1651399656

This comment is still correct - if we provide a client wrapper then we mark the "native" proto client as deprecated, but still export it in case users have a need to work with the native client. At some point we can probably stop exporting native APIs, or more ideally starting using a different proto generator for JS...

RE: https://github.com/viamrobotics/viam-typescript-sdk/pull/316/files#r1653127091

We do want to export the custom type since it allows users to write type annotations. However, we generally convert it to a POJO with `.AsObject`. The whole idea is that users don't have to deal with native proto JS, just objects.

zaporter-work: Good point @maximpertsov . Added `asObject` in d27e425

maximpertsov: also any new types needs to be exported in `src/main.ts` to actually be importable from the package

zaporter-work: Oh -- my newness to this is showing -- I had assumed an export * above this. 7c16a9a

maximpertsov: no it's fine, this is an all-too-common oversight

zaporter-work: Are we worried that changing to return the `asObject` version is a breaking change?

maximpertsov: ah that's annoying, we were inconsistent with this type. @njooma I defer to you on this one - how do you feel about making this breaking change?

purplenicole730: (Commenting so I can see what decision we make)
From what I know, it's a pretty well-established pattern in the TS SDK that the responses are not the raw proto type but the objects. Any users who used this method had to treat this response differently than any other. But this means that future users will have to do this as well if we decide not to fix this.

njooma: I think we should make this breaking change to bring things in line with all the other objects. Let's just make sure to announce it appropriately.

As for renaming -- I think we can rename to `ArmJointPositions` and, moving forward, make that the default -- if we're re-exporting a proto from a specific package, prefix with the package name. (So "common" protos wouldn't have a prefix, but specific component ones would).

zaporter-work: @njooma So just to confirm: you are okay with this PR moving forward as-is? If so, would you be willing to approve?

(I'm happy to make the eng-breaking-changes announcement unless you'd rather do it.)

maximpertsov (approved):

purplenicole730 (approved):

njooma (approved):

PR #327, opened 2024-07-29
Add the ability to abort an initial connection attempt

ethanlook:
I discovered this while working with an offline robot - if you're making your initial connection with retries, there was previously no ability to abort the reconnect attempts once they started. This allows the user to pass in an object with the rest of the dial options that they can later cancel.

## Change log

- Add `reconnectAbort` option to `DialOpts`
- Do not retry if `reconnectAbort.abort` is true
- Update `vanilla` example with usage

## Tested

- Vanilla example
- Used in app, observed behavior is resolved!

### DIFF
--- a/examples/vanilla/package-lock.json
+++ b/examples/vanilla/package-lock.json
@@ -19,7 +19,7 @@
     },
     "../..": {
       "name": "@viamrobotics/sdk",
-      "version": "0.16.0",
+      "version": "0.19.1",
       "license": "Apache-2.0",
       "dependencies": {
         "@viamrobotics/rpc": "^0.2.5",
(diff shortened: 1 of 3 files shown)


ethanlook:
> 1. Why does it have to be an object? Can it not just be a boolean? Does it have something to do with reference semantics if it's an object?

Exactly, we want the reference semantics of the object so that the connection can be cancelled later. Initially, the caller would set `reconnectAbort.abort` to `false` and then later decide to stop attempting reconnects and set to `true`.

> The connection logic returns a promise, right? Is there a way to have the cancellation of the promise work as the abort signal?

Unfortunately promises don't support cancel natively. The closest thing is you can use an [`AbortController`](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) with `fetch` but that doesn't help us with the way we establish transport - we would have to manually listen to the signal just like we're doing here.

ethanlook:
@njooma Happy to change the name! I chose this one because it matches the `exponential-backoff` parameter. `reconnectAbortSignal` sounds good to me!

ethanlook:
@njooma renamed to `reconnectAbortSignal`: [c8d93fb](https://github.com/viamrobotics/viam-typescript-sdk/pull/327/commits/c8d93fb8a8e32552844ba400b2832520184596c7)

njooma (approved):

PR #334, opened 2024-08-05
APP-4023: Add wrapper for RobotService.restartModule

michaellee1019:


src/robot/client.ts
```diff
@@ -813,4 +813,23 @@ export class RobotClient extends EventDispatcher implements Robot {
...
+      request.setModuleId(moduleId);
+    }
+    if (moduleName !== undefined) {
+      request.setModuleName(moduleName);
+    }
+    const response = await promisify<
+      proto.RestartModuleRequest,
+      proto.RestartModuleResponse
+    >(robotService.restartModule.bind(robotService), request);
+    return response.toObject();
```
njooma, line 833: Probably don't need to actually return anything since the return type doesn't include any data

njooma (changes requested):

src/robot/client.ts
michaellee1019: Yeah makes sense. If we add fields to the response in the future, then we can return it.

stuqdog (changes requested):
Can we add a stub for the `restartModule` method to `robot/robot.ts` with a docstring that explains the method a bit (what it does, if/when it raises, behavior when neither param is provided, stuff like that)?

michaellee1019:
> Can we add a stub for the `restartModule` method to `robot/robot.ts`

Ah, didn't realize this existed. Will do!

stuqdog (approved):

njooma (approved):

PR #342, opened 2024-08-27
Return Uint8Array image from captureAllFromCamera

micheal-parks:
### Overview

Currently captureAllFromCamera returns a base64 encoded image from `captureAllFromCamera`. This PR updates the method implementation so that a more efficient `Uint8Array` is returned instead.

### DIFF
--- a/src/services/vision/client.ts
+++ b/src/services/vision/client.ts
@@ -189,8 +189,17 @@ export class VisionClient implements Vision {
       pb.CaptureAllFromCameraRequest,
       pb.CaptureAllFromCameraResponse
     >(service.captureAllFromCamera.bind(service), request);
+
+    const image = response.getImage();
+
     return {
-      image: response.getImage()?.toObject(),
+      image: image
+        ? {
+            format: image.getFormat(),
+            sourceName: image.getSourceName(),
+            image: image.getImage_asU8(),
+          }
+        : undefined,
       classifications: response
         .getClassificationsList()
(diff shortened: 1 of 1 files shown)


src/services/vision/client.ts
```diff
@@ -189,8 +189,17 @@ export class VisionClient implements Vision {
...
+
+    const image = response.getImage();
+
     return {
-      image: response.getImage()?.toObject(),
+      image: image
+        ? {
+            format: image.getFormat(),
+            sourceName: image.getSourceName(),
+            image: image.getImage_asU8(),
```
ethanlook, line 200: Is the return value of this function strictly typed? If so, all good. If not, this image response could drift if someone changed the image definition (e.g. added a new field to `Image`)

ethanlook (approved):

src/services/vision/client.ts
micheal-parks: It is, it's following the definition of `Image` and a return type error will be thrown if it doesn't adhere to it.

micheal-parks:
Just to add a little extra clarification: this is correcting a problem with the original implementation and aligning behavior with the go and python SDKs. Additionally this will help performance with a part of app that a potential customer is looking at. @stuqdog happen to know who would be a good reviewer on your team? Thanks in advance for any help!

njooma (approved):

micheal-parks:
Thanks @njooma!

PR #344, opened 2024-08-29
Add `getSessions` Wrapper to RobotClient

zaporter-work:
Imported into app and tested that the type signatures are correct (have not manually run it yet)

Used `proto.Session.AsObject[]` instead of `proto.Session[]` because I've seen a lot of the api wrappers move to the .AsObject form. 

However, `GetOperations` still returns the `proto.Operation[]` form, so for consistency also happy to swap over to that.

### DIFF
--- a/src/robot/client.ts
+++ b/src/robot/client.ts
@@ -631,6 +631,18 @@ export class RobotClient extends EventDispatcher implements Robot {
     }
   }
 
+  // SESSIONS
+
+  async getSessions() {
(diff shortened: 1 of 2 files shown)


zaporter-work:
(Added more reviewers in case people are out. Would be nice  to get this into this cycle's release)

njooma (approved):
`asObject` is good!

maximpertsov (approved):

PR #351, opened 2024-09-13
RSDK-8156: binaryDataCaptureUpload and tabularDataCaptureUpload actually do require dataRequestTimes

gloriacai01:
-remove optionality of `dataRequestTime` parameter in `binaryDataCaptureUpload` and `tabularDataCaptureUpload `
-had to switch ordering of `tags` and `dataRequestTime` parameters as a required parameter cannot come after an optional parameter and `tags` is optional

### DIFF
--- a/src/app/data-client.test.ts
+++ b/src/app/data-client.test.ts
@@ -1036,8 +1036,8 @@ describe('DataSyncClient tests', () => {
         componentType,
         componentName,
         methodName,
-        tags,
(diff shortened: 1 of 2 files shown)


njooma (approved):
Nice! This prior to merging, a notification needs to be sent in the breaking changes slack channel so that all the teams are aware this is coming

purplenicole730 (approved):

PR #365, opened 2024-09-24
Merge in pre connect-es rpc/js

edaniels:
Discussed with @mcous and @ethanlook, we're going to merge the rpc/js into this repo before the connect-es changes in order to ease the transition as far as build systems go. Later PRs will bring the connect-es code in and then later, the examples and gh actions.

Lint disables you see applied here are done because it's not worth fixing code that is about to be replaced. It works as is.

### DIFF
--- a/src/app/viam-transport.ts
+++ b/src/app/viam-transport.ts
@@ -1,5 +1,5 @@
 import { grpc } from '@improbable-eng/grpc-web';
-import { dialDirect } from '../rpc/dial';
+import { dialDirect } from '../rpc';
 
 import {
   AuthenticateRequest,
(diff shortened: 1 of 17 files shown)


src/app/viam-transport.ts
```diff
@@ -1,5 +1,5 @@
 import { grpc } from '@improbable-eng/grpc-web';
-import { dialDirect } from '@viamrobotics/rpc';
+import { dialDirect } from '../rpc/dial';
```
ethanlook, line 2: What do you think about adding an `rpc/index.ts` to simplify imports?

ethanlook (approved):
Awesome, thanks! This definitely will reduce the burden of rpc changes in the TS SDK for the time being.

src/app/viam-transport.ts
edaniels: sure!

njooma (approved):
I'm assuming the `rpc` directory is basically just copied from goutils and will trust that it functions the same

PR #368, opened 2024-09-27
Migrate SDK to connect-es

edaniels:
I need to check some integrations with app but this ready to review.

### Major Changes
* connect-es is in use now which breaks every usage of proto. e.g. pbObj.getSomeField() is now pbObj.someField
* AsObject gets removed everywhere
* I unified Credentials into a single type since it had two definitions across rpc/js and the sdk
  *  Someone will need to update app examples. Alternatively, I can change the type back to have authEntity be separate. Someone let me know!
* React Native has many more polyfills in order to satisfy the modern nature of connect-es, sadly. We could make this simpler in the future probably.

### Minor Fixes
* exponential-backoff now works as expected. It was a classic setting a property to undefined makes it present/set.

This PR is unfortunately huge but everything works and is tested across every example. Happy to go over anything in person though or if there's a better way to split it up, just holler. But... it's more deletions than additions and **everything** should be easier to read.


doc/dependency_decisions.yml
```diff
@@ -27,22 +27,13 @@
...
-    :when: 2023-05-01 18:11:37.804857463 Z
-- - :license
-  - '@improbable-eng/grpc-web-fake-transport'
-  - Apache 2.0
-  - :who:
-    :why: 'TODO(RSDK-583): why is license finder unable to detect this license automatically?'
-    :license_links: https://www.npmjs.com/package/@improbable-eng/grpc-web-fake-transport
-    :versions: []
-    :when: 2023-05-01 18:11:37.804857463 Z
+- - :approve
```
edaniels, line 30: Need approval on this but it's apache 2 and bsd-3

Makefile
```diff
@@ -31,7 +31,7 @@ test-watch: $(node_modules) build-buf
 lint: $(node_modules) build-buf
 	npm run lint
 	npm run typecheck
-	npm run check
+	npm run check -- --ignore=@bufbuild/protobuf
```
edaniels, line 34: the new major version (2) is incompatible with other packages used by connect-es

njooma (approved):
Not going through every line of code but did some tests and it all works so 👍

PR #384, opened 2024-10-09
RSDK-6171-add-client-api-for-debugging-grpc-calls

purplenicole730:
Working example:
```
  await motor1Client.setPower(0.5);
  let opts = VIAM.enableDebugLogging('this-is-nicole');
  await motor1Client.setPower(0.75, {}, opts);
  motor1Client.callOptions = VIAM.enableDebugLogging('motor-opts');
  await motor1Client.setPower(1);
  await motor1Client.setPower(0.25, {}, opts);
  VIAM.disableDebugLogging(motor1Client.callOptions);
  await motor1Client.setPower(0);
  await motor1Client.setPower(0.3, {}, opts);
```
> debug rdk.resource_manager.rdk:component:motor/motor1 fake/motor.go:212 Motor SetPower 0.300000 traceKey this-is-nicole
> debug rdk.resource_manager.rdk:component:motor/motor1 fake/motor.go:212 Motor SetPower 0.250000 traceKey this-is-nicole
> debug rdk.resource_manager.rdk:component:motor/motor1 fake/motor.go:212 Motor SetPower 1.000000 traceKey motor-opts
> debug rdk.resource_manager.rdk:component:motor/motor1 fake/motor.go:212 Motor SetPower 0.750000 traceKey this-is-nicole


src/components/base/base.ts
```diff
@@ -51,7 +51,7 @@ export interface Base extends Resource {
   stop(extra?: Struct): Promise<void>;
 
   /** Return true if the base is in motion. */
-  isMoving(extra?: Struct): Promise<boolean>;
+  isMoving(): Promise<boolean>;
```
purplenicole730, line 54: This is technically a breaking change. None of the other `isMoving` functions have an `extra` parameter, so I thought to remove it. Let me know if I should not do this.

src/utils.ts
```diff
@@ -31,10 +32,57 @@ export const doCommandFromClient = async function doCommandFromClient(
...
+  return finalOpts;
+};
+
+export const deleteMetadata = (opts: CallOptions, key: string): void => {
+  const { [key]: _, ...remainingHeaders } = opts.headers as Record<
+    string,
+    string
+  >;
+  opts.headers = remainingHeaders;
+};
```
njooma, line 88: Are these helpers used? Or just for potential users ?

purplenicole730: These two helpers not being used right now, but I thought it would be good to include them for potential usage. This was also introduced in the Python SDK, so I thought it would be a good idea to keep things similar. Should I remove it?

stuqdog: I think we should leave it in personally. We're now offering users access to passing metadata so we might as well make it easy for them to do so.

stuqdog (approved):

src/utils.ts
njooma: we can leave in. i was just wondering if i missed something

njooma (approved):

PR #386, opened 2024-10-10
Fix ice candidate proto to webrtc interface conversion

edaniels:


.eslintrc.cjs
```diff
@@ -30,6 +30,12 @@ module.exports = {
     '@typescript-eslint/no-explicit-any': 'warn',
     '@typescript-eslint/prefer-promise-reject-errors': 'warn',
     'unicorn/prefer-add-event-listener': 'warn',
+    '@typescript-eslint/strict-boolean-expressions': [
```
edaniels, line 33: Turning this on with the exception of booleans feels very worth it to me. I've written numerous bugs due to how optionals are checked in a truthy/falsey way.

src/rpc/signaling-exchange.ts
```diff
@@ -365,30 +366,30 @@ const iceCandidateFromProto = (i: ICECandidate) => {
   const candidate: RTCIceCandidateInit = {
     candidate: i.candidate,
   };
-  if (i.sdpMid) {
+  if (i.sdpMid !== undefined) {
```
edaniels, line 369: This is the fix. We we were excluding the right properties because some of them were falsey

src/rpc/signaling-exchange.ts
```diff
@@ -242,8 +242,9 @@ export class SignalingExchange {
     try {
       await this.pc.addIceCandidate(cand);
     } catch (error) {
+      console.log('error adding ice candidate', error); // eslint-disable-line no-console
```
edaniels, line 245: This will report an error if this ever happens again and bubble up to the reason of dial failure instead of undefined.

njooma (approved):

ethanlook (approved):
Did not review code but I did try it in a deployed temp environment and it works! https://github.com/viamrobotics/app/pull/6514

PR #390, opened 2024-10-15
APP-6639 - DoCommand from control tab for services no longer works (regression)

zaporter-work:
Ticket: https://viam.atlassian.net/browse/APP-6639

We imported the wrong Generic here.

I have built this and manually imported into app & confirmed that this fixed the issue.


ethanlook:
Can we bump the TS SDK version in this PR so we can bump in App and RC as well? We've gotten a couple reports of folks running into this and I'd like to avoid a rollback of the most recent bump that caused this.

https://viam.atlassian.net/browse/APP-6639

njooma:
I'll merge this and create a release asap

PR #391, opened 2024-10-15
RSDK-8989 - Always provide error on base channel close

edaniels:
@zaporter-work fwiw, this has been a bug prior to the new connect-es migration

Screenshot 2024-10-15 at 11 12 45 AM

### DIFF
--- a/src/rpc/base-channel.ts
+++ b/src/rpc/base-channel.ts
@@ -40,10 +40,6 @@ export class BaseChannel {
     });
   }
 
-  public close() {
-    this.closeWithReason(undefined);
-  }
-
   public isClosed() {
     return this.closed;
   }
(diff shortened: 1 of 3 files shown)


zaporter-work (approved):
Interesting -- must not have had the requisite conditions to trigger it. Don't remember ever seeing this in the console. Could go check an old PR dev env.

Thanks for the quick fix!

njooma (approved):

njooma (approved):
Do these methods need to be public? Since RPC is inside the SDK now, there aren't any external users of this.

If we can make these private/internal, then we don't have to worry about this stuff being breaking

edaniels:
> Do these methods need to be public? Since RPC is inside the SDK now, there aren't any external users of this.
> 
> If we can make these private/internal, then we don't have to worry about this stuff being breaking

It's on the base channel which isn't to be used by external users (it's not exported) but it's composed by another class, so it needs to be public

PR #393, opened 2024-10-16
docs: update Node.js example to run as expected

HipsterBrown:
I'm very excited to be able to use the Viam TypeScript SDK for my Node.js projects! Thanks for all the hard work to make this integration possible @edaniels and @njooma! :clap:

After running through the current documentation and example application, I ran into a few issues that I'm resolving with this PR.

### DIFF
--- a/Node.md
+++ b/Node.md
@@ -1,10 +1,10 @@
-# Node & Viam's TypeScript SDK
+# Node.js & Viam's TypeScript SDK
 
-This document contains detailed instructions on including Viam in your Node project. For a runnable example, see the [examples directory](/examples/node/).
+This document contains detailed instructions on including Viam in your Node.js project. For a runnable example, see the [examples directory](/examples/node/).
 
 ## Requirements
 
-This document assumes you already have Node installed. If not, follow the [instructions](https://nodejs.org/en/learn/getting-started/how-to-install-nodejs) provided by Node.
+This document assumes you already have Node.js >= version 20 installed. If not, follow the [instructions](https://nodejs.org/en/learn/getting-started/how-to-install-nodejs) provided by Node.js.
 
 ### Dependencies
 
@@ -19,12 +19,12 @@ You can use either Yarn or NPM to install the dependencies. This document will u
 
 ### Polyfills
 
-Using the SDK with Node also requires the use of some polyfills. In your application's entrypoint (`main.ts`, `index.ts`, or something similar), you will need to register those polyfills:
+Using the SDK with Node.js also requires the use of some polyfills. In your application's entrypoint (`main.ts`, `index.ts`, or something similar), you will need to register those polyfills:
 
 ```ts
 // main.ts
 
-import wrtc = require('node-datachannel/polyfill');
+const wrtc = require('node-datachannel/polyfill');
 for (const key in wrtc) {
   (global as any)[key] = (wrtc as any)[key];
 }
@@ -32,12 +32,12 @@ for (const key in wrtc) {
 
 ### Transport
 
-Communicating with your Viam machine in Node requires the use of a custom transport. In your app's entrypoint, you will also need to register the custom transport:
+Communicating with your Viam machine in Node.js requires the use of a custom transport. In your app's entrypoint, you will also need to register the custom transport:
 
 ```ts
 // main.ts
 
-import connectNode = require('@connectrpc/connect-node');
+const connectNode = require('@connectrpc/connect-node');
 globalThis.VIAM = {
   GRPC_TRANSPORT_FACTORY: (opts: any) =>
     connectNode.createGrpcTransport({ httpVersion: '2', ...opts }),
@@ -51,9 +51,9 @@ To use the SDK, you can use similar instructions to those found on the [document
 ```ts
 // main.ts
 
-import VIAM = require('@viamrobotics/sdk');
-import wrtc = require('node-datachannel/polyfill');
-import connectNode = require('@connectrpc/connect-node');
+const VIAM = require('@viamrobotics/sdk');
+const wrtc = require('node-datachannel/polyfill');
+const connectNode = require('@connectrpc/connect-node');
 globalThis.VIAM = {
   GRPC_TRANSPORT_FACTORY: (opts: any) =>
     connectNode.createGrpcTransport({ httpVersion: '2', ...opts }),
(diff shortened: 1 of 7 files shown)


Node.md
```diff
@@ -1,10 +1,10 @@
-# Node & Viam's TypeScript SDK
+# Node.js & Viam's TypeScript SDK
```
HipsterBrown, line 1: Using the official name "Node.js" for consistency across the documentation.

Node.md
```diff
@@ -1,10 +1,10 @@
...
 
-This document contains detailed instructions on including Viam in your Node project. For a runnable example, see the [examples directory](/examples/node/).
+This document contains detailed instructions on including Viam in your Node.js project. For a runnable example, see the [examples directory](/examples/node/).
 
 ## Requirements
 
-This document assumes you already have Node installed. If not, follow the [instructions](https://nodejs.org/en/learn/getting-started/how-to-install-nodejs) provided by Node.
+This document assumes you already have Node.js >= version 20 installed. If not, follow the [instructions](https://nodejs.org/en/learn/getting-started/how-to-install-nodejs) provided by Node.js.
```
HipsterBrown, line 7: The `--env-file` flag was introduced in v20, so calling this out just in case some folks are running v18 which is still under long term support.

Node.md
```diff
@@ -19,25 +19,25 @@ You can use either Yarn or NPM to install the dependencies. This document will u
...
 ### Polyfills
 
-Using the SDK with Node also requires the use of some polyfills. In your application's entrypoint (`main.ts`, `index.ts`, or something similar), you will need to register those polyfills:
+Using the SDK with Node.js also requires the use of some polyfills. In your application's entrypoint (`main.ts`, `index.ts`, or something similar), you will need to register those polyfills:
 
 ```ts
 // main.ts
 
-import wrtc = require('node-datachannel/polyfill');
+const wrtc = require('node-datachannel/polyfill');
```
HipsterBrown, line 27: Node.js supports two module formats: CommonJS (`require`) and ESM (`import`). While they can be mixed together in some programs, the expected syntax for each format is different. If we wanted to use ESM import statements it would look a little like reversed Python:

```ts
import wrtc from 'node-datachannel/polyfill';
```

However after updating the example code to use ESM, it appears the Viam TypeScript SDK does not have an ESM compatible module export for Node.js at the moment. So I've updated the example code to use CommonJS for now.

examples/node/package.json
```diff
@@ -17,7 +17,6 @@
     "ts-protoc-gen": "^0.14.0",
     "typescript": "^5.2.2"
   },
-  "type": "module",
```
HipsterBrown, line 20: This field enforces ESM for Node.js projects, which is not currently possible with the TS SDK, so removing this to run the project as CommonJS.

examples/node/src/main.ts
```diff
@@ -1,6 +1,8 @@
-import VIAM = require('@viamrobotics/sdk');
-import wrtc = require('node-datachannel/polyfill');
-import connectNode = require('@connectrpc/connect-node');
+const VIAM = require('@viamrobotics/sdk');
+const wrtc = require('node-datachannel/polyfill');
+const connectNode = require('@connectrpc/connect-node');
+
+// @ts-expect-error
```
HipsterBrown, line 5: This is required to pass the TypeScript compiler (`tsc`). If the project was updated to use something like [`tsx`](https://tsx.is/node-enhancement#swap-node-for-tsx), it could run in a single command: `tsx  --env-file=.env src/main.ts`

Otherwise, once Node.js v22 is more widely used, it could be run directly without the extra dependency using the [type stripping feature](https://nodejs.org/en/learn/typescript/run-natively): `node --experimental-strip-types --env-file=.env src/main.ts`

examples/node/src/main.ts
```diff
@@ -12,7 +14,7 @@ for (const key in wrtc) {
 async function connect() {
   const host = process.env.HOST;
   const apiKeyId = process.env.API_KEY_ID;
-  const apiKeySecret = process.env.API_KEY_SECRET;
+  const apiKeySecret = process.env.API_KEY;
```
HipsterBrown, line 17: The README uses `API_KEY` in the `.env` file instead of `API_KEY_SECRET` and this naming is more consistent across the other SDK documentation.

Node.md
edaniels: This file works fine for me in Node 20 mod your API key secret fix

examples/node/package.json
edaniels: same for node 20

edaniels (commented):
This all looks good to me other than a discussion for main.ts (mod api key secret fix). 20 is about a year old and 18 is EOL in 6 months. I think it'd be better to be on 20.

Node.md
HipsterBrown: `import wrtc = require('node-datachannel/polyfill')` works in Node 20? That's quite odd since it doesn't work in the REPL:

image

Could `tsc` be transpiling the code to be valid?

edaniels: Yes tsc or tsx would transpile to js since node can't run typescript.

njooma: I believe eslint/the community in general prefers ESM style imports in TS? At least i often see lint errors when i do `const foo = require('bar')`

HipsterBrown: Node 22 can run TypeScript with the experimental type stripping flag and would fail with the example code as is. I think we should have valid JS in the examples, even if tsc papers over the syntax errors.

HipsterBrown: > I believe eslint/the community in general prefers ESM style imports in TS? At least i often see lint errors when i do const foo = require('bar')

Yes, ESM is the official standard and should be used when possible. However, the current build output of the TS SDK cannot be imported as ESM in Node.js:

image

(The above example in the REPL is using the dynamic ESM import syntax but the same error applies when running `main.ts` with static ESM imports)

HipsterBrown:
I don't know why the E2E tests are failing. I ran them locally with the same result.

edaniels:
@HipsterBrown maybe it's best to move to tsx instead of tsc while getting `node --experimental-strip-types src/main.ts` to work too? Whatever gives the most compat I'm all for

e2e/tests/connect-and-status.spec.ts
```diff
@@ -7,12 +7,12 @@ test('check resource names and multiple statuses', async ({ page }) => {
   const resourceNames = page
     .getByTestId('resource-names')
     .getByTestId('resource-name');
-  await expect(resourceNames).toHaveCount(4);
+  await expect(resourceNames).toHaveCount(3);
```
HipsterBrown, line 10: I'm not sure why the number of builtin resources is only 1 now, maybe something changed in the server implementation. It appears to be unrelated to my changes.

edaniels (approved):

njooma (approved):

PR #401, opened 2024-10-30
RSDK-9147: Change Typescript TabularDataBySQL/MQL return type to raw BSON

jckras:
For context: The return type of the TabularDataBySQL/MQL proto has changed to return both a list of structs, the existing return type, and also a list of bytearrays that represent BSON.

Added a required dependency to bsonfy to use the `BSON` library
BSON provides a deserialize() function for converting lists of byte sequences into lists of javascript objects.

**Testing:**
Updated the `data` testing input to include `Date` objects, as the Typescript SDK will receive date fields in this format. I added a check to ensure that `Date` inputs are returned as native typescript `Date` objects

I ensured that date fields are returned as native typescript Date objects instead of strings by running tabularDataByMQL against real tabular data on viam-dev and printed the decoded data and their types. As you can see below the date and time elements are of type "Date" (click on the photo for clarity)
Screenshot 2024-10-31 at 1 33 08 PM

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,6 +12,7 @@
         "@bufbuild/protobuf": "^1.10.0",
         "@connectrpc/connect": "^1.6.0",
         "@connectrpc/connect-web": "^1.6.0",
+        "bsonfy": "^1.0.2",
         "exponential-backoff": "^3.1.1"
       },
       "devDependencies": {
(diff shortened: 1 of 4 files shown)


purplenicole730 (approved):

purplenicole730:
Similarly to the other SDKs, is this a breaking change?

jckras:
> Is this a breaking change?
Yes! we talked off-line about this but yes since the dates/times will be given back as native Date objects and not strings this is a breaking change

stuqdog (approved):

PR #407, opened 2024-11-06
Fix resource name in motion tests

njooma:


jckras (approved):
LGTM - just out of curiosity what is the localName property used for in the context of resource naming?

njooma:
> LGTM - just out of curiosity what is the localName property used for in the context of resource naming?

Honestly a good question. I'm not sure what the new fields are. We'll have to chat with the Netcode team

PR #413, opened 2024-11-12
Add type: module to package.json

zaporter-work:
The teleop team is trying to switch our usage of TabularDataByMQL to use the ts-sdk wrapper (so that we can use a stable function but allow the data team to change the api as-desired)

Ran into:

```
ui start: (node:95053) Warning: To load an ES module, set "type": "module" in the package.json or use the .mjs extension.
ui start: (Use `node --trace-warnings ...` to show where the warning was created)
ui start: 12:53:00 PM [vite] Error when evaluating SSR module /src/lib/api/client.ts: failed to import "@viamrobotics/sdk"
ui start: |- /home/zack/work/app/node_modules/.pnpm/@viamrobotics+sdk@0.28.0/node_modules/@viamrobotics/sdk/dist/main.es.js:22354
ui start: export {
ui start: ^^^^^^
ui start: SyntaxError: Unexpected token 'export'
```


package.json
```diff
@@ -1,10 +1,11 @@
 {
   "name": "@viamrobotics/sdk",
-  "version": "0.29.0",
+  "version": "0.29.1",
```
zaporter-work, line 3: Ok to minor bump?

package.json
```diff
@@ -16,7 +17,7 @@
     }
   },
   "scripts": {
-    "prebuild": "ts-node ./scripts/write-versions.ts",
+    "prebuild": "tsx ./scripts/write-versions.ts",
```
zaporter-work, line 20: ts-node struggles to run when package.json has `type: module`. Ran into this: 0962f516cb74b715d01434b21a295d27d7b09c60

Decided it was easier to swap to tsx https://www.npmjs.com/package/tsx

I might be missing something small... I didn't want to mess with the ts-config for the entire project just for this helper script.

package.json
njooma: We don't update the version manually but rather through a workflow when it's time to release, so I would change this back please

mcous: Switching from CJS to ESM should be considered a breaking change, since it will prevent existing CJS consumers from importing the package. Since this package is `0.*` (`<1.0.0`), a bump to `0.30.0` satisfies semver requirements for communicating a break

package.json
njooma: Ya probably better to use tsx? Seems like everything I can find online about it is positive

njooma (changes requested):

package.json
zaporter-work: SGTM. Done in ad6e215

package.json
zaporter-work: @njooma Do you expect to have other more complex scripts that would benefit from types? If so, might be worth keeping tsx. If not, agreed that typescript isn't providing much value.

njooma: Let's strip it for now and if we add more complex scripts then we can deal with it then. no need to keep it around for a "maybe" if it's entirely unused now (aside from this use case)

zaporter-work: Swapped to .js in 2fd1e38

mcous (approved):

package.json
mcous: Just realized I misspoke a little here: we've already got a CJS package export defined, so probably this won't break any consumers as written.

The being said, I've seen other libraries consider touching type and exports to be breaking changes, so I think erring on the side of caution here and assuming potential downstream breakage is a good idea

package.json
```diff
@@ -5,6 +5,7 @@
   "main": "./dist/main.umd.js",
   "module": "./dist/main.es.js",
   "types": "./dist/main.d.ts",
+  "type": "module",
```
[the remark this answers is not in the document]
zaporter-work: This has been a fun deep-dive to read into. I had no idea exports were sorted. Thanks for thinking about this!

> we'll want to put the require condition above the import condition, to ensure conditions are most-to-least specific. 

Reading https://nodejs.org/api/packages.html#conditional-exports , it seems like the order of `import` and `require` is not important because they are entirely mutually exclusive. Am I missing a case where it would be ambiguous which export to use? Though, I like the idea of switching to `require -> default` (seems like a good change)

This dual-packaging example from nodejs matches our use case: https://github.com/nodejs/package-examples?tab=readme-ov-file#approach-2-isolate-state

They have the following `package.json` example:
```json
{
  "type": "module",
  "exports": {
    "import": "./index.mjs",
    "require": "./index.cjs"
  }
}
```

> We could also consider dropping the CJS export and UMD build to make things easier on ourselves

I like the idea. The dual-packaging docs mention that the global state of the cjs module is not shared with the es module -- which would interact poorly with the ts-sdks extensive use of globals. (though, admittedly, I think it is unlikely that people are exporting cjs libraries using the ts-sdk)


-- 

Deferring to the sdk team because they know more about consumption of this package.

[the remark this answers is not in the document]
zaporter-work: I've spent the last 20 minutes researching cjs vs esm in 2024 and I can't find any solid evidence that cjs is actually deprecated / dying (despite my strong bias towards esm). Multiple github issues threads on the nodejs forms indicate strongly that cjs isn't going anywhere and is still a first-class packaging format. 

So to revisit the original comment:
> We could also consider dropping the CJS export and UMD build to make things easier on ourselves

Are we actually running into difficulty with the current CJS & UMD setup? Is it worth dropping and potentially having to re-add if some customer needs to use legacy code?

I suspect the answer to both of these is a slight no. Happy to be wrong if people feel strongly.

---

I have swapped `import -> require` to `require -> default` in 2fe46c8 . I have imported this into the app and verified that everything works.

At this point, if it is okay with everyone involved, I would like to move forward with this PR mostly as-is (unless there are potentially backwards-incompatible issues).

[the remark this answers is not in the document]
njooma: Agreed, I don't think any difficulty we currently face is worth dropping CJS for now.

PR #418, opened 2024-11-22
[RSDK-9251] - Add options helpers to stream client

seanavery:
## Description
This PR adds a few helpers to interact with dynamic resolution stream server endpoints.

## Tests
Added success and failure test cases for the new stream interface methods.


seanavery:
It looks like the `npm check` is returning 1 in CI... Would it be ok to bump these dependencies?

```
> @viamrobotics/sdk@0.30.0 check
> npm-check --production --ignore=@bufbuild/protobuf

- Checking for unused packages. --skip-unused if you don't want this.
- Checking npm registries for updated packages.

@connectrpc/connect       😎  MAJOR UP  Major update available. https://github.com/connectrpc/connect-es#readme
                                       npm install @connectrpc/connect@2.0.0 --save to go from 1.6.1 to 2.0.0

@connectrpc/connect-web   😎  MAJOR UP  Major update available. https://github.com/connectrpc/connect-es#readme
                                       npm install @connectrpc/connect-web@2.0.0 --save to go from 1.6.1 to 2.0.0

Use npm-check -u for interactive update.
make: *** [Makefile:34: lint] Error 1
```

src/extra/stream/client.ts
```diff
@@ -83,6 +86,84 @@ export class StreamClient extends EventDispatcher implements Stream {
...
+   */
+  async getOptions(resourceName: string): Promise<Resolution[]> {
+    const fetchOptions = async (name: string): Promise<Resolution[]> => {
+      const request = new GetStreamOptionsRequest({ name });
+      this.options.requestLogger?.(request);
+      try {
+        const response = await this.client.getStreamOptions(request);
+        return response.resolutions;
+      } catch {
+        return [];
```
jckras, line 105: Why return an empty list when the available resolution is not available and not an error message?

src/extra/stream/client.ts
```diff
@@ -83,6 +86,84 @@ export class StreamClient extends EventDispatcher implements Stream {
...
+  /**
+   * Set the livestream options for a camera component. This will change the
+   * resolution of the stream to the specified width and height.
+   *
+   * @param name - The name of a camera component.
+   * @param width - The width of the resolution.
+   * @param height - The height of the resolution.
+   */
+  async setOptions(name: string, width: number, height: number) {
+    const request = new SetStreamOptionsRequest({
```
jckras, line 128: do we need any error checking that the width/height parameters are set within a specific bound?

src/extra/stream/client.ts
```diff
@@ -83,6 +86,84 @@ export class StreamClient extends EventDispatcher implements Stream {
...
+      resolution: {
+        width,
+        height,
+      },
+    });
+    this.options.requestLogger?.(request);
+    try {
+      await this.client.setStreamOptions(request);
+    } catch {
+      // Try again with just the resource name
```
jckras, line 139: why do we try again with just the resource name? if getValidSDPTrackName(name) is false then we are just falling back to using an unformatted name?

jckras (approved):
LGTM.  One thing I’m wondering is whether we should add validation for input parameters across the methods (ex: resourceName/name should be non-empty strings, width/height in setOptions should be positive numbers) other than that just left a few questions for my own understanding

src/extra/stream/client.ts
seanavery: Yeah good question! I toyed around with erroring out in the failure case but thought it would be nicer to just return empty array to avoid complexity when calling it in the frontend and trying to fill a dropdown or such with the return. (This is presumptuous as I do not have that much experience with frontend stuff) Happy to change the behavior to throw if that is better.

src/extra/stream/client.ts
seanavery: All of the GRPC field checks should happen in the server endpoint and it should hopefully error out here with the correct message. Would it be good to add additional checks here in the client before sending?

src/extra/stream/client.ts
seanavery: Not sure... Was initially thrown off by this but this is the behavior from the other stream client endpionts so I copied it. If we want I can do a follow up PR to test this out and remove across the stream client if we dont need it.

seanavery:
@stuqdog I unsuccessfully attempted bump on `connectrpc` dependencies, this caused a dependency conflict on `protobuf`, so needed to bump both.
```
    "@bufbuild/protobuf": "^2.2.0",
    "@connectrpc/connect": "^2.0.0",
    "@connectrpc/connect-web": "^2.0.0",
```

Immediately getting errors from API change when trying to `make clean` then `make build`
```> @viamrobotics/sdk@0.29.0 build:bundle
> vite build

The CJS build of Vite's Node API is deprecated. See https://vitejs.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated for more details.
vite v5.4.8 building for production...
✓ 308 modules transformed.
x Build failed in 263ms
error during build:
src/gen/common/v1/common_pb.js (8:24): "proto3" is not exported by "node_modules/@bufbuild/protobuf/dist/esm/index.js", imported by "src/gen/common/v1/common_pb.js".
file: /Users/sean/viam-typescript-sdk/src/gen/common/v1/common_pb.js:8:24

 6: // @ts-nocheck
 7: 
 8: import { MethodOptions, proto3, Struct, Timestamp, Value } from "@bufbuild/protobuf";
                            ^
 9: 
10: /**

    at getRollupError (file:///Users/sean/viam-typescript-sdk/node_modules/rollup/dist/es/shared/parseAst.js:392:41)
    at error (file:///Users/sean/viam-typescript-sdk/node_modules/rollup/dist/es/shared/parseAst.js:388:42)
    at Module.error (file:///Users/sean/viam-typescript-sdk/node_modules/rollup/dist/es/shared/node-entry.js:14844:16)
    at Module.traceVariable (file:///Users/sean/viam-typescript-sdk/node_modules/rollup/dist/es/shared/node-entry.js:15291:29)
    at ModuleScope.findVariable (file:///Users/sean/viam-typescript-sdk/node_modules/rollup/dist/es/shared/node-entry.js:13192:39)
    at MemberExpression.bind (file:///Users/sean/viam-typescript-sdk/node_modules/rollup/dist/es/shared/node-entry.js:6860:49)
    at CallExpression.bind (file:///Users/sean/viam-typescript-sdk/node_modules/rollup/dist/es/shared/node-entry.js:2836:23)
    at CallExpression.bind (file:///Users/sean/viam-typescript-sdk/node_modules/rollup/dist/es/shared/node-entry.js:11148:15)
    at VariableDeclarator.bind (file:///Users/sean/viam-typescript-sdk/node_modules/rollup/dist/es/shared/node-entry.js:2836:23)
    at VariableDeclaration.bind (file:///Users/sean/viam-typescript-sdk/node_modules/rollup/dist/es/shared/node-entry.js:2832:28)
make: *** [build-js] Error 1
```

Is it ok to merge a PR with a ❌ test? Happy to create a ticket for this with results from my bump attempt.

src/extra/stream/client.ts
seanavery: Looks like the flutter sdk only tries the sdp track name.
https://github.com/viamrobotics/viam-flutter-sdk/blob/main/lib/src/media/stream/client.dart#L88-L97

src/extra/stream/client.ts
jckras: I think simple is better! keep how it is

src/extra/stream/client.ts
jckras: ah got it. if the checks are happening elsewhere then we def don't need to duplicate it

src/extra/stream/client.ts
jckras: Unfortunately I don't know enough on whether one way is better than another here but I think stick with consistency here on the other stream clients?

PR #419, opened 2024-11-27
DATA-3396: Add GetLatestTabularData endpoint to TypeScript SDK

vpandiarajan20:
We added a new Data endpoint GetLatestTabularData https://github.com/viamrobotics/app/pull/6896, this PR adds the endpoint to Typescript SDK.

### DIFF
--- a/src/app/data-client.ts
+++ b/src/app/data-client.ts
@@ -699,7 +699,7 @@ export class DataClient {
    *   the data
    * @param methodName The data capture method name
    * @returns A tuple containing [timeCaptured, timeSynced, payload] or null if
-   *   no data has been synced for the specified resource or the most recently
+   *   no data has been synced for the specified resource OR the most recently
    *   captured data was over a year ago
    */
   async getLatestTabularData(


src/app/data-client.ts
```diff
@@ -687,6 +687,43 @@ export class DataClient {
...
+   * @param methodName The data capture method name
+   * @returns A tuple containing [timeCaptured, timeSynced, payload] or null if
+   *   no data has been synced
+   */
+  async getLatestTabularData(
+    partId: string,
+    resourceName: string,
+    resourceSubtype: string,
+    methodName: string
+  ) {
```
kaywux, line 709: Function header needs a return type?

kaywux (changes requested):

kaywux (approved):
Nice, LGTM!

njooma (approved):

src/app/data-client.ts
```diff
@@ -687,6 +687,44 @@ export class DataClient {
...
+   *   the data
+   * @param methodName The data capture method name
+   * @returns A tuple containing [timeCaptured, timeSynced, payload] or null if
+   *   no data has been synced for the specified resource OR the most recently
+   *   captured data was over a year ago
+   */
+  async getLatestTabularData(
+    partId: string,
+    resourceName: string,
+    resourceSubtype: string,
```
njooma, line 708: How would y'all feel about swapping these 2? I kinda like the idea of progressive specificity -- go from least specific (part) to most specific (method name). 

So the order would be part -> subtype -> name -> method

kaywux: I think if you guys don't mind, we'd like to keep the order as is, at least in this PR - this follows the existing ordering convention for data that's shown in the UI, among other places. We're already exposing this ordering to users (below is a screenshot from the image / sensor view's filter panel) so would like to keep things consistent.

Screenshot 2024-12-03 at 9 53 18 AM

Of course, happy to discuss further on this - the field order switch could definitely be a larger change across data

PR #428, opened 2024-12-16
DATA-3444 Add exportTabularData to data client

katiepeters:
This PR adds the new Data endpoint [ExportTabularData](https://github.com/viamrobotics/app/pull/6969/files) to the typescript SDK.

### DIFF
--- a/src/app/data-client.ts
+++ b/src/app/data-client.ts
@@ -38,6 +38,22 @@ interface TabularData {
   timeReceived?: Date;
 }
 
+interface TabularDataPoint {
+  partId: string;
+  resourceName: string;
+  resourceSubtype: string;
+  methodName: string;
+  timeCaptured: Date;
+  organizationId: string;
+  locationId: string;
+  robotName: string;
+  robotId: string;
+  partName: string;
+  methodParameters: JsonValue;
+  tags: string[];
+  payload: JsonValue;
+}
+
 export type Dataset = Partial<PBDataset> & {
   created?: Date;
 };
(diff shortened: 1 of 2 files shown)


src/app/data-client.ts
```diff
@@ -38,6 +38,22 @@ interface TabularData {
   timeReceived?: Date;
 }
 
+interface TabularDataPoint {
+  partId: string;
+  resourceName: string;
+  resourceSubtype: string;
+  methodName: string;
+  timeCaptured: Date;
```
stuqdog, line 46: per failing tests, I think this needs to be `Date | undefined` since it's possible no `timeCaptured` exists on the response.

stuqdog (approved):
The failing test needs to be addressed, otherwise lgtm!

src/app/data-client.ts
```diff
@@ -53,6 +69,68 @@ export class DataClient {
...
+   * @returns An array of unified tabular data and metadata.
+   */
+  async exportTabularData(
+    partId: string,
+    resourceName: string,
+    resourceSubtype: string,
+    methodName: string,
+    startTime?: Date,
+    endTime?: Date
+  ) {
```
katiepeters, line 92: @stuqdog Wondering if I should change this to return `AsyncIterableIterator<TabularDataPoint>`. Thoughts?

src/app/data-client.ts
katiepeters: I didn't realize we had to account for that with protobuf! Noted

src/app/data-client.ts
stuqdog: Hmm... I suspect we need to be `await`ing ourselves either way to convert into a `TabularDataPoint` (or get into some funky async functional mapping). I guess that might be useful if someone just wants to stream all tabular data as it continues to come in? That seems like it adds more complexity/work to users so my thought is no, but there may be a use case I'm not thinking about (could someone use that to tail all data as it comes in? is that a use case we expect to be common?).

tl;dr weakly I support keeping as-is but open to being convinced otherwise.

katiepeters: I'm thinking of the use case where someone wants to retrieve a lot of data. Possibly even download it as a file

PR #430, opened 2024-12-17
DATA-3444 (part 2) Add non-null assertions

katiepeters:
Because time captured, method parameters, and payload will always be included in the response, I wanted to propose adding non-null assertions


stuqdog (approved):

purplenicole730 (approved):
learned something new (`!`)

PR #444, opened 2025-01-17
APP-7392: Add getVersion to robot client

ethanlookpotts:
I need `getVersion` so that I can report RDK version to Sentry for control tab errors. This adds the RPC to the robot client as passthrough.

Let me know what you think - I can also just call `machine.robotService.getVersion({})` directly, so happy to close this PR.


stuqdog (approved):
looks reasonable to me! I see no harm in exposing this helper method.

PR #451, opened 2025-01-27
APP-7497: Add Button interface and client

ethanlookpotts:
This adds the new Button API type to the TypeScript SDK using the following API: https://github.com/viamrobotics/api/blob/main/proto/viam/component/button/v1/button.proto

This will get used in the new control cards to render a button card and call the Push API.

## Change log

- Bump API version
- Add Button interface
- Add ButtonClient implementation
- Export Button, ButtonClient, and protos from the package

### DIFF
--- a/src/components/button/client.ts
+++ b/src/components/button/client.ts
@@ -1,9 +1,7 @@
 import { Struct, type JsonValue } from '@bufbuild/protobuf';
 import type { CallOptions, PromiseClient } from '@connectrpc/connect';
 import { ButtonService } from '../../gen/component/button/v1/button_connect';
-import {
-  PushRequest,
-} from '../../gen/component/button/v1/button_pb';
+import { PushRequest } from '../../gen/component/button/v1/button_pb';
 import type { RobotClient } from '../../robot';
 import type { Options } from '../../types';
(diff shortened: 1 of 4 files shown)


src/components/button/client.ts
```diff
@@ -0,0 +1,50 @@
...
+  public callOptions: CallOptions = { headers: {} as Record<string, string> };
+
+  constructor(client: RobotClient, name: string, options: Options = {}) {
+    this.client = client.createServiceClient(ButtonService);
+    this.name = name;
+    this.options = options;
+  }
+
+  async press(extra = {}, callOptions = this.callOptions) {
+    const request = new PushRequest({
```
njooma, line 28: Why call the client method `press` when the API method is `push`?

ethanlookpotts: Lol whoops - I blame Claude/Cursor 🤣 thanks for the catch

ethanlookpotts: [6de507b](https://github.com/viamrobotics/viam-typescript-sdk/pull/451/commits/6de507b91a7755b1eedf78585959ef992220ff67)

njooma (approved):

stuqdog (approved):

PR #466, opened 2025-02-11
DATA-3757 Update TabularDataByMQL to have optional bool for use_recent_data

vijayvuyyuru:


stuqdog (commented):
This all looks reasonable to me though we should figure out why the linter step is failing before merging.

vijayvuyyuru:
@stuqdog Looks like a dependency needed to be updated. Fixed now

stuqdog (approved):

PR #469, opened 2025-02-12
Add ml model service

DTCurrie:
We have an incoming request to render the response of the `Metadata()` in the test controls for the ML Model Service card. This exposes the `metadata` API to be used.


src/services/ml-model/ml-model.ts
```diff
@@ -0,0 +1,12 @@
+import type { Struct } from '@bufbuild/protobuf';
+import type {
+  FlatTensors,
+  InferResponse,
+  MetadataResponse,
+} from '../../gen/service/mlmodel/v1/mlmodel_pb';
```
zaporter-work, line 6: We should re-export these types (like arm/arm.ts) so that other clients (svelte-sdk) can use them

zaporter-work (approved):
LGTM (have not manually tested this)

src/services/ml-model/ml-model.ts
DTCurrie: Ah yeah good call.

bhaney (approved):

micheal-parks (approved):

njooma (approved):
thanks!

src/services/ml-model/index.ts
```diff
@@ -0,0 +1,7 @@
+export type { MLModel } from './ml-model';
+export { MLModelClient } from './client';
+export type {
+  FlatTensors,
+  InferResponse,
+  MetadataResponse,
+} from '../../gen/service/mlmodel/v1/mlmodel_pb';
```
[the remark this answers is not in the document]
DTCurrie: That makes sense, I will update them to shadow types. I also may just reduce it to the `FlatTensors` shape, since they other are response shapes but `FlatTensors` is used as an argument for one of the APIs, so I could see users wanting to pre-construct that before passing it along.

PR #472, opened 2025-02-18
Make client name field public

micheal-parks:
### Overview

It's often useful to read the corresponding resource name from a client, for example when executing requests on the `robotClient` in a scope where only a resource client and not the original resource name is available. In this PR I've switched all names to public but kept the readonly modifier.

### DIFF
--- a/src/components/arm/client.ts
+++ b/src/components/arm/client.ts
@@ -22,7 +22,7 @@ import type { Arm } from './arm';
  */
 export class ArmClient implements Arm {
   private client: PromiseClient<typeof ArmService>;
-  private readonly name: string;
+  public readonly name: string;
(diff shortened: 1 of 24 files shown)


zaporter-work (approved):
I'm in favor of this change -- svelte-sdk has to pair clients and their names together. If this info was on the client, it would simplify our code. 

Are there any downsides to this?

micheal-parks:
I don't see any. It's still readonly, and the user knows the name of the client when passing it to the constructor. This is useful metadata that should be made available IMO. In non-ts-land people can already change this value since all modifiers are compiled away.

stuqdog (approved):
seems reasonable to me!

PR #475, opened 2025-02-26
Add actuating methods

njooma:
clone of https://github.com/viamrobotics/viam-python-sdk/pull/857


stuqdog (commented):
Code looks reasonable to me! I wonder if it's not a bit safer to invert the logic here though? It's a bit more annoying for us to maintain but if a new method is added and we completely forget whether or not to consider the sessions aspect of it, it seems to me that it's safer to assume that we _should_ have sessions and risk annoying/noisy error messages rather than assume that we _shouldn't_ have sessions and risk that an actuating process continues after connectivity is lost.

njooma:
Agreed -- see note in the [python SDK PR](https://github.com/viamrobotics/viam-python-sdk/pull/857) as to why I did it this way instead of the inverse

stuqdog (approved):

purplenicole730 (approved):

stuqdog (approved):

PR #477, opened 2025-03-04
[RSDK-10123] user defined metadata crud apis

JosephBorodach:
# Description
Add user-defined metadata CRUD operations 

**See ticket [here](https://viam.atlassian.net/browse/RSDK-10123)**

## Todo
- [ ] Do I need to manually test?


src/app/app-client.ts
```diff
@@ -1175,4 +1183,100 @@ export class AppClient {
...
+    return this.client.updateLocationMetadata({ locationId, data });
+  }
+
+  /**
+   * Retrieves user-defined metadata for a robot.
+   *
+   * @param id The ID of the robot
+   * @returns The metadata associated with the robot
+   */
+  async getRobotMetadata(id: string): Promise<GetRobotMetadataResponse> {
```
njooma, line 1241: You mix/match prepending the type of object to `Id`: `locationId` and `organizationId`, but just `id` for `robot` and `robotPart`. Let's try to be consistent? Even if the proto has different names, we can make the user-facing API consistent.

I don't mind one way or the other, maybe a slight preference to just `id` everywhere since the object is already specified in the method name, but I'm never against hyper-specificity if you want to do `robotId`

njooma: Which actually brings up another question -- do we want to call it `robot` or `machine`?

njooma (changes requested):

src/app/app-client.ts
JosephBorodach: Thanks for calling this out - I agree we should be consistent, let's go with `id`!

My gut tells me robot because that's what we called it on the protos end of things, but what do you think?

src/app/app-client.ts
```diff
@@ -1175,4 +1183,100 @@ export class AppClient {
...
+  /**
+   * Updates user-defined metadata for an organization.
+   *
+   * @param id The ID of the organization
+   * @param data The metadata to update
+   * @returns Response indicating success or failure
+   */
+  async updateOrganizationMetadata(
+    id: string, 
+    data: Record<string, any>
```
JosephBorodach, line 1206: [q] @njooma - the linter seems to be getting upset at the usage of `any`, do you know of an alternative that I could use?

JosephBorodach: Is it ok for me to use `// eslint-disable-next-line  @typescript-eslint/no-explicit-any
`?

src/app/app-client.ts
njooma: I want to ask @JessamyT here for docs perspective: should we be using `robot` or `machine`? Most other functions are still using `robot`, so do we try for consistency, or do we start the migration to `machine`?

src/app/app-client.ts
njooma: Hmm.... I'm confused by this functionality actually? Looking at the proto, I see that the request type is `map<string, google.protobuf.Any>`, and `google.protobuf.Any` "contains an arbitrary serialized protocol buffer message along with a URL that describes the type of the serialized message." (from the [docs](https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/any.proto#L42)).

So from what I can see, I think the `data` parameter in TS should also be `google.protobuf.Any`? Or if we want to be an arbitrary type `any` (like string/number/whatever), the proto might need to be changed?

src/app/app-client.ts
JessamyT: My uninformed inclination is to migrate towards `machine` since we do use that everywhere throughout the docs, even when talking about SDKs that use `robot`. Like we basically say `robot = machine` all over the place.

src/app/app-client.ts
JosephBorodach: I think the former option makes sense - is that feasible?

JosephBorodach: I pushed what I think uses `google.protobuf.Any` - does this look good to you?

src/app/app-client.ts
```diff
@@ -1175,4 +1184,104 @@ export class AppClient {
...
+  ): Promise<GetOrganizationMetadataResponse> {
+    return this.client.getOrganizationMetadata({ organizationId: id });
+  }
+
+  /**
+   * Updates user-defined metadata for an organization.
+   *
+   * @param id The ID of the organization
+   * @param data The metadata to update
+   * @returns Response indicating success or failure
```
stuqdog, line 1205: I think we should get rid of this and just make this a `void`-returning method. The response doesn't have anything in it so it doesn't actually tell the user anything useful, but that isn't necessarily obvious to a user so they might think that there's like a `success` field in the response or something. And in case of a failure, there would be a panic or raised error, so the `response` wouldn't be received.

src/app/app-client.ts
```diff
@@ -1175,4 +1184,104 @@ export class AppClient {
...
+  async getLocationMetadata(id: string): Promise<GetLocationMetadataResponse> {
+    return this.client.getLocationMetadata({ locationId: id });
+  }
+
+  /**
+   * Updates user-defined metadata for a location.
+   *
+   * @param id The ID of the location
+   * @param data The metadata to update
+   * @returns Response indicating success or failure
```
stuqdog, line 1229: See above re: making this void-returning.

src/app/app-client.ts
```diff
@@ -1175,4 +1184,104 @@ export class AppClient {
...
+  async getMachineMetadata(id: string): Promise<GetRobotMetadataResponse> {
+    return this.client.getRobotMetadata({ id });
+  }
+
+  /**
+   * Updates user-defined metadata for a robot.
+   *
+   * @param id The ID of the machine
+   * @param data The metadata to update
+   * @returns Response indicating success or failure
```
stuqdog, line 1253: See above re: making this void-returning.

src/app/app-client.ts
```diff
@@ -1175,4 +1184,104 @@ export class AppClient {
...
+  ): Promise<GetRobotPartMetadataResponse> {
+    return this.client.getRobotPartMetadata({ id });
+  }
+
+  /**
+   * Updates user-defined metadata for a machine part.
+   *
+   * @param id The ID of the machine part
+   * @param data The metadata to update
+   * @returns Response indicating success or failure
```
stuqdog, line 1279: See above re: making this void-returning.

stuqdog (changes requested):
A small suggestion, otherwise looks good!

stuqdog (approved):

src/app/app-client.ts
```diff
@@ -1175,4 +1180,100 @@ export class AppClient {
...
+  /**
+   * Retrieves user-defined metadata for an organization.
+   *
+   * @param id The ID of the organization
+   * @returns The metadata associated with the organization
+   */
+  async getOrganizationMetadata(
+    id: string
+  ): Promise<GetOrganizationMetadataResponse> {
+    return this.client.getOrganizationMetadata({ organizationId: id });
```
njooma, line 1193: We should return a dictionary here. Since the rpc returns a `Struct`, we should be able to convert that `Struct` into a native JS obj to make it easier for users to work it

src/app/app-client.ts
```diff
@@ -1175,4 +1180,100 @@ export class AppClient {
...
+
+  /**
+   * Updates user-defined metadata for an organization.
+   *
+   * @param id The ID of the organization
+   * @param data The metadata to update
+   */
+  async updateOrganizationMetadata(
+    id: string,
+    data: Record<string, Any>
```
njooma, line 1204: Similarly, we should take in a record of string to lowercase `any` (the typescript type), and then convert that into a `pb.Struct` so that it's easier for the user to use

njooma (changes requested):
The two comments I left are applicable to all methods added

edobranov (changes requested):
(left [one comment](https://github.com/viamrobotics/viam-typescript-sdk/pull/477#discussion_r1987979788) that I want to make sure we're clear on, otherwise nothing that's blocking from me)

src/app/app-client.ts
njooma: I can see that being useful. particularly when searching for functions, you could try to do a find for `robot` and see all the functions available on a robot. @JessamyT can we call this `robot` and then to a formal migration when we're ready to rename everything to `machine`?

JessamyT: Sounds good to me!

JosephBorodach: @njooma @JessamyT - is it ok with you if we stick with `robot` in this pr and we cut a ticket to migrate everything over to `machine` at some point? I do not feel strongly one way or the other

src/app/app-client.ts
JosephBorodach: Do things look good as I have it now? I am having a really hard time getting the unit tests to pass (not too familiar with typescript) - do I need to add unit tests?

src/app/app-client.ts
```diff
@@ -109,6 +112,27 @@ export const createPermission = (
...
 
+/**
+ * Decodes a map<string, google.protobuf.Any> into a
+ * record of plain JavaScript objects keyed by the same map keys.
+ *
+ * @param data A record of string -> google.protobuf.Any
+ * @returns A record whose keys match `data`,
+ *          with values unpacked from Struct messages
+ */
+export const decodeMetadataMap = (data: Record<string, Any>): Record<string, JsonValue> => {
```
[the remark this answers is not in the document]
JosephBorodach: Nope! thanks for clarifying

src/app/app-client.ts
```diff
@@ -1175,4 +1199,125 @@ export class AppClient {
...
+   * @param data The metadata to update
+   */
+  async updateOrganizationMetadata(
+    id: string,
+    data: Record<string, JsonValue>
+  ): Promise<void> {
+    const convertedData = new Struct({ fields: {} });
+    for (const [key, val] of Object.entries(data)) {
+      convertedData.fields[key] = Value.fromJson(val);
+    }
```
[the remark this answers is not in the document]
JosephBorodach: yes, thanks!

src/app/app-client.ts
```diff
@@ -1175,4 +1176,125 @@ export class AppClient {
...
+    const jsonResponse = response.toJson() as {
+      data?: Record<string, JsonValue>;
+    };
+    return jsonResponse.data ?? {};
+  }
+
+  /**
+   * Updates user-defined metadata for a machine robot.
+   *
+   * @param id The ID of the machine robot
```
edobranov, line 1288:
```suggestion
   * Updates user-defined metadata for a robot part.
   *
   * @param id The ID of the robot part
```

src/app/app-client.ts
```diff
@@ -1175,4 +1176,125 @@ export class AppClient {
   ): Promise<CreateKeyFromExistingKeyAuthorizationsResponse> {
     return this.client.createKeyFromExistingKeyAuthorizations({ id });
   }
+
+  /**
+   * Retrieves user-defined metadata for an organization.
+   *
+   * @param id The ID of the organization
+   * @returns The metadata associated with the organization as a plain JS object
```
edobranov, line 1184:
```suggestion
   * @returns The metadata associated with the organization
```
The other docstrings don't mention this so I think we can remove it

edobranov (approved):
LGTM % a couple docstring tweaks

src/app/app-client.ts
JosephBorodach: Thanks for calling this out!

njooma (approved):
Nice!! Thanks for all the updates. 

For future reference, you could also do something like this so that you aren't doing the same cast over and over again:

`return (response.data?.toJson() as JsonObject) ?? {};`

PR #486, opened 2025-03-10
DATA-3861 Update TS SDK to use new binary data id fields

vijayvuyyuru:
This PR changes the TS SDK to use the new id field. The change can be [read about more here.](https://docs.google.com/document/d/1zapFPyrkIRNrnD7wyD9XolOVUKFgth2cXQRoX53jifU/edit?usp=sharing) but basically, we're deprecating the BinaryID struct and replacing it with a string. This points all methods to these new fields.

### DIFF
--- a/src/app/data-client.spec.ts
+++ b/src/app/data-client.spec.ts
@@ -17,7 +17,6 @@ import {
   BinaryDataByFilterResponse,
   BinaryDataByIDsRequest,
   BinaryDataByIDsResponse,
-  BinaryID,
   BoundingBoxLabelsByFilterRequest,
   BoundingBoxLabelsByFilterResponse,
   CaptureInterval,
--- a/src/app/data-client.spec.ts
+++ b/src/app/data-client.spec.ts
@@ -17,6 +17,7 @@ import {
   BinaryDataByFilterResponse,
   BinaryDataByIDsRequest,
   BinaryDataByIDsResponse,
+  BinaryID,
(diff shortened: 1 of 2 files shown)


stuqdog (approved):
looks reasonable to me! I assume we'll be making some sort of breaking change announcement?

Also it looks like linter is complaining about an unused import, so probably should fix that before we merge.

vijayvuyyuru:
@stuqdog Yup! The corresponding BE change hasnt been merged yet so I won't merge this for at least a couple days and will make an announcement before. Thanks!

njooma (changes requested):
This is a breaking change, so it would probably be helpful to allow users to input either `BinaryID` or `string` (you can use union types in TypeScript), and then check within the body of the method if you got a `string` or `BinaryID` and appropriately apply that in the request.

UNLESS `BinaryIDs` will just no longer work, in which case this is fine.

vijayvuyyuru:
@njooma I didn't know that feature existed. I can update this to do that, one question tho is that feature isn't possible in every language. Doing this would result in divergent SDK experiences where some are broken after this change and others aren't. Is that fine?

njooma (approved):

njooma (approved):

njooma:
@vijayvuyyuru can this be merged?

vijayvuyyuru:
@njooma Yep, sorry was on vacation and then waited a day to make sure the app changes this corresponds do don't break anything.

PR #502, opened 2025-03-25
DOCS-3540: Add camera samples

npentrel:


src/components/camera/camera.ts
```diff
@@ -43,12 +65,33 @@ export interface Camera extends Resource {
    *
    * @param mimeType - A specific MIME type to request. This is not necessarily
    *   the same type that will be returned.
+   *
+   * @example
+   * ```ts
+   * const camera = new VIAM.CameraClient(machine, 'my_camera');
+   * const mimeType = 'image/jpeg';
+   * const image = await camera.renderFrame(mimeType);
```
npentrel, line 73: This returns a blob. It would be good if we had a better sample but I'm unsure how to turn the blob into something usable. Do you happen to have a sample?

njooma: I do not, but you can probably display a blog by setting it on an html `img` tag (untested)

njooma (approved):

src/components/camera/camera.ts
npentrel: I shall leave it to the reader for now. It seems working with blobs is a common enough thing

PR #503, opened 2025-03-25
DOCS-3548: Add vision service samples

npentrel:


src/services/vision/vision.ts
```diff
@@ -26,6 +33,22 @@ export interface Vision extends Resource {
...
+   * ```ts
+   * const camera = new VIAM.CameraClient(machine, 'my_camera');
+   * const vision = new VIAM.VisionClient(machine, 'my_vision');
+   *
+   * const mimeType = 'image/jpeg';
+   * const image = await camera.getImage(mimeType);
+   * const detections = await vision.getDetections(
+   *   image,
+   *   width,
+   *   height,
```
njooma, line 47: Where are `width` and `height` coming from?

njooma (commented):
Just one question

src/services/vision/vision.ts
npentrel, line 47:
```suggestion
   *   600,
   *   600,
```

npentrel: good catch

njooma (approved):

PR #508, opened 2025-03-27
DOCS-3537: Add data code samples

npentrel:


njooma (changes requested):
Can you make sure this is up to date with main? This PR still uses `BinaryID` in a lot of places, and that has been deprecated

npentrel:
Updated. It seems those changes merged the same day I requested review on this.

njooma (approved):

PR #515, opened 2025-04-01
DOCS-3539: Add code samples for arm, base, board

JessamyT:


src/components/arm/arm.ts
```diff
@@ -10,22 +10,60 @@ export const { JointPositions: ArmJointPositions } = armApi;
...
 export interface Arm extends Resource {
-  /** Get the position of the end of the arm expressed as a pose */
+  /** Get the position of the end of the arm expressed as a pose
+   *
+   * @example
+   *
+   * ```ts
+   * const arm = new VIAM.ArmClient(machine, 'my_arm');
+   * const pose = await arm.getEndPosition();
+   * console.log(pose);
```
npentrel, line 20: Please remove all the logs since we don't have them for other languages either

JessamyT: Okay--we do have print statements for some like 
`print(f"Pin {tick.pin_name} changed to {'high' if tick.high else 'low'} at {tick.time}")`
so I'll leave a couple in where analogous to those and remove the rest.

njooma (approved):

PR #537, opened 2025-04-14
DOCS-3790: Add docs backlinks for input controller-switch and vision

JessamyT:


njooma (approved):
I'm approving with the assumption that the line breaks in the comments don't actually affect the links

JessamyT:
The linter made me do it :D I staged and they seem to work.

PR #540, opened 2025-04-15
make name public and export types for rendering

DTCurrie:
As I was working on the MLModel service test card, I noticed we were having some type issues with the client. First, the `name` field needed to be public. Second, we were missing some type exports that are pretty necessary to render anything from the metadata response. Shouldn't be anything breaking.

### DIFF
--- a/src/services/ml-model/client.ts
+++ b/src/services/ml-model/client.ts
@@ -10,8 +10,9 @@ import { MLModelService } from '../../gen/service/mlmodel/v1/mlmodel_connect';
 
 export class MLModelClient implements MLModel {
   private client: PromiseClient<typeof MLModelService>;
-  private readonly name: string;
   private readonly options: Options;
+
+  public readonly name: string;
   public callOptions: CallOptions = { headers: {} as Record<string, string> };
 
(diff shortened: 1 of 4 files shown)


micheal-parks (approved):

stuqdog (approved):

purplenicole730 (approved):

src/services/ml-model/client.ts
```diff
@@ -10,8 +10,9 @@ import { MLModelService } from '../../gen/service/mlmodel/v1/mlmodel_connect';
 
 export class MLModelClient implements MLModel {
   private client: PromiseClient<typeof MLModelService>;
-  private readonly name: string;
```
njooma, line 13: Interesting. Not sure why it was private to begin with

njooma (approved):

src/services/ml-model/client.ts
DTCurrie: Normally this would be caught because we would extend `Resource` which requires `name` to be public. But the `MLModelService` does not implement the `doCommand`, which is a requirement for a `Resource`. So the type-checker didn't yell at me when I missed it last time.

PR #543, opened 2025-04-16
DOCS-3790: Add motion and SLAM backlinks

JessamyT:
I noticed that the getInternalState and getPointCloudMap examples are not rendering in the [TS docs](https://ts.viam.dev/classes/SlamClient.html), and this seemed to be fixed by changing the arrow functions to regular functions. However I'm not sure whether this creates other problems.

### DIFF
--- a/src/services/motion/motion.ts
+++ b/src/services/motion/motion.ts
@@ -57,6 +57,9 @@ export interface Motion extends Resource {
    * const moved = await motion.move(goalPoseInFrame, gripperName);
    * ```
    *
+   * For more information, see [Motion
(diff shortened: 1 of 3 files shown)


stuqdog (approved):
LGTM. I'm not aware of any issues caused by changing from arrow functions, but I am not a TS expert.

JessamyT:
I'm definitely not either, and I'm just not sure why these two functions were defined this way in the first place. Is there a TS expert who could confirm for sure? Or is there a test that would've failed if I had broken something?

njooma:
I think as all the other clients don't use arrow functions, these shouldn't either

PR #551, opened 2025-04-30
DATA-4054 Data pipelines Typescript support

n0nick:
This PR:

1. Adds SDK coverage for all new DataPipelinesService APIs.
2. Replaces UseRecentData usage with the new TabularDataSource format (DATA-3966).

Manual testing: See https://gist.github.com/n0nick/52bb4954472e618d302bd20a350712b9


src/app/data-client.ts
```diff
@@ -219,22 +231,37 @@ export class DataClient {
...
+    // Legacy support for useRecentData, which is now deprecated.
+    let dataSource = tabularDataSource;
+    if (
+      useRecentData &&
+      (!dataSource || dataSource.type === TabularDataSourceType.UNSPECIFIED)
+    ) {
+      dataSource = new TabularDataSource({
+        type: TabularDataSourceType.HOT_STORAGE,
+      });
+    }
```
n0nick, line 259: @vijayvuyyuru in the context of [DATA-4091](https://viam.atlassian.net/browse/DATA-4091) i think this is safe right? if dataSource was null before, we will still send null

vijayvuyyuru: Im not 100% sure, I think if tabularDataSource is undefined, doing dataSource.type might cause an exception?

katiepeters: If we're handling this in app, what's the reason behind adding it here?

src/app/data-client.ts
```diff
@@ -1315,6 +1342,216 @@ export class DataClient {
...
+   * ```ts
+   * const page = await dataClient.listDataPipelineRuns(
+   *   '123abc45-1234-5678-90ab-cdef12345678'
+   * );
+   * const nextPage = await page.nextPage();
+   * ```
+   *
+   * @returns A page of data pipeline runs
+   */
+  async nextPage(): Promise<ListDataPipelineRunsPage> {
```
katiepeters, line 1530: Cool!

njooma (approved):
minor nit, otherwise good!

katiepeters (approved):

PR #563, opened 2025-05-21
Emit dial events rather than logging them

micheal-parks:
Currently we don't allow end-users to consume dial events, we only log them to the JS console. Logging isn't a good idea for production apps because:

1. It could pose security risks for clients that don't want to expose reconnect logic.
2. In cases where we want to make connection logic transparent to the user the JS console is often hidden away.

This PR improves both situations by instead having the RobotClient emit dial events, which we can present in UIs, and removing logs.


src/events.ts
```diff
@@ -36,7 +37,7 @@ export class EventDispatcher {
     this.listeners[type]?.delete(listener);
   }
```
DTCurrie, line 38: Why aren't we passing `<T>` to `once`, `has`, and `off`?

DTCurrie (approved):

src/events.ts
micheal-parks: Added to `once`, can't think of any type safety added by the others can you?

src/robot/client.ts
```diff
@@ -234,12 +235,14 @@ export class RobotClient extends EventDispatcher implements Robot {
     }
     void backOff(async () => this.connect(), backOffOpts)
       .then(() => {
-        // eslint-disable-next-line no-console
-        console.debug('Reconnected successfully!');
+        this.emit(MachineConnectionEvent.DIAL_EVENT, {
+          message: 'Reconnected successfully',
```
njooma, line 239: Why is this a `DIAL_EVENT` and not `CONNECTED`?

src/robot/dial.ts
```diff
@@ -115,14 +118,19 @@ const dialWebRTC = async (conf: DialWebRTCConf): Promise<RobotClient> => {
...
+
   await client.connect({
     priority: conf.priority,
     dialTimeout: conf.dialTimeout ?? DIAL_TIMEOUT,
     creds: conf.credentials,
   });
 
-  // eslint-disable-next-line no-console
-  console.debug('connected via WebRTC');
+  client.emit(MachineConnectionEvent.DIAL_EVENT, {
```
njooma, line 131: Similarly, why is this a `DIAL_EVENT` and not `CONNECTED`?

src/events.ts
```diff
@@ -9,21 +9,22 @@ export enum MachineConnectionEvent {
   CONNECTED = 'connected',
   DISCONNECTING = 'disconnecting',
   DISCONNECTED = 'disconnected',
+  DIAL_EVENT = 'dialing',
```
njooma, line 12: Is `DIALING` different than `CONNECTING`?

njooma (commented):
A few questions

src/events.ts
DTCurrie: No, `once` make sense.

PR #579, opened 2025-06-26
[APP-8686] Add GetAppBranding API

JosephBorodach:
Ticket is [here](https://viam.atlassian.net/browse/APP-8686)

This adds the GetAppBranding() function for the Viam single page apps project


JosephBorodach:
@njooma would you mind taking a look when you get a chance? It should be quick review

PR #619, opened 2025-08-25
Move dial logic into `RobotClient` class

micheal-parks:
### Overview

Currently the only way to connect to a robot is by calling `createRobotClient`. This function is async, creating a `RobotClient` class and attempting connection before returning the client. This is problematic for the following reasons:

* Aborting the initial connection is much more difficult and error prone than disconnecting after the first successful connection because the client instance is not made available. Currently the only way to abort the initial connection attempt is by passing an object, named an "abort controller" (which is not a true [AbortController](https://developer.mozilla.org/en-US/docs/Web/API/AbortController)), and mutating it. This method is fragile and can easily break under typical immutability patterns.
* It's impossible to emit specific dialing events to surface to users because the logic is in plain functions and are not methods of an EventDispatcher class.

### Solution

This PR moves all of the dial logic into the RobotClient class. This is `not` a breaking change, the behavior of `createRobotClient` and `RobotClient` is the same. However, it is now possible to instantiate a robotClient before dialing, making disconnection much easier:

```ts
const client = new RobotClient()

client.dial(config).then(() => {
  console.log("connected")
})

// This will disconnect during the first connection attempt.
client.disconnect()
```

Additionally, since the dial logic is now within the RobotClient class, we can emit events that can be piped to a UI or console to provide users more understanding of what's currently happening:

```ts
client.on('dialing', (event) => {
  console.log(event.method) // webrtc or grpc
  console.log(event.attempt) // attempt number
})
```


src/robot/client.ts
```diff
@@ -70,16 +118,70 @@ export interface ConnectOptions {
   dialTimeout?: number;
 }
 
+export const isDialWebRTCConf = (value: DialConf): value is DialWebRTCConf => {
```
mattmacf98, line 121: [q] is there a reason we don't use a library like zod for parsing and validating here? is it too large to include in the sdk?

src/robot/client.ts
```diff
@@ -435,6 +559,118 @@ export class RobotClient extends EventDispatcher implements Robot {
...
+
+    this.sessionOptions.disabled = conf.disableSessions ?? false;
+
+    await this.connect({
+      priority: conf.priority,
+      dialTimeout: conf.dialTimeout ?? DIAL_TIMEOUT,
+      creds: conf.credentials,
+    });
+
+    return this;
```
mattmacf98, line 586: [q] why do we need to return this here? would the caller already have a reference to the object since they are calling this method on it?

micheal-parks: We don't need to, here I'm just following a common class convention to return the object to allow chaining:

```
const client = await new RobotClient().dial()
```

I'm not tied to it so can remove if others don't like this

mattmacf98: EDIT: nvm I see this was already the code in place and this was just moving it

src/robot/client.ts
```diff
@@ -151,17 +253,32 @@ export class RobotClient extends EventDispatcher implements Robot {
 
   private slamServiceClient: PromiseClient<typeof SLAMService> | undefined;
 
+  private currentRetryAttempt = 0;
+
   constructor(
-    serviceHost: string,
```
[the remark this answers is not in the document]
micheal-parks: Yep, totally agreed, I tried to make as few changes in this PR, but would love to do more cleanup later

DTCurrie (approved):

src/robot/client.ts
```diff
@@ -151,17 +253,32 @@ export class RobotClient extends EventDispatcher implements Robot {
...
   ) {
     super();
-    this.serviceHost = serviceHost;
-    this.webrtcOptions = webrtcOptions;
-    this.directOptions = directOptions;
-    this.sessionOptions = sessionOptions;
+
+    if (serviceHost !== undefined) {
+      this.serviceHost = serviceHost;
+    }
```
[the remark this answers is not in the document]
micheal-parks: Making it optional was a middle ground for not making any breaking changes to `createRobotClient`, but I don't see any reactive app passing it as a constructor arg since it's awkward and is part of the dial config anyway. With react, svelte, etc, it makes much more sense to pass it during the dial call:

```ts
const client = new RobotClient()

React.useEffect(() => {
  client.dial(config)
  return () => client.disconnect()
}, [config])
```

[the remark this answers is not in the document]
micheal-parks: If we make it required, a lot of apps will just end up passing an empty string to the constructor.

PR #621, opened 2025-08-27
RSDK-10408 Add world-state-store service

DTCurrie:
Adds the implementation for the `WorldStateStoreService`.

API PR: https://github.com/viamrobotics/api/pull/695
RDK changes: https://github.com/viamrobotics/rdk/pull/5149

Scope doc: https://docs.google.com/document/d/1ionvyBa7x3HZU_rwDPBvrAUrQjBrP6sJ10Z_xbBTue8/edit?tab=t.0#heading=h.tcicyojyqi6c


src/services/world-state-store/client.ts
```diff
@@ -0,0 +1,95 @@
...
+  private readonly options: Options;
+  public callOptions: CallOptions = { headers: {} as Record<string, string> };
+
+  constructor(client: RobotClient, name: string, options: Options = {}) {
+    this.client = client.createServiceClient(WorldStateStoreService);
+    this.name = name;
+    this.options = options;
+  }
+
+  async listUUIDs(extra = {}, callOptions = this.callOptions) {
```
micheal-parks, line 31: For all of these functions we should convert the bytes uuids to string before they hit the user. Maybe use something like this? https://github.com/domske/uuid-tool

DTCurrie: I updated to include a specific new type and utility function for handling this. Since we are just interacting with UUIDs here I wanted to keep it scoped pretty tightly, but we can always elevate it if we think this is useful for other resources.

micheal-parks (approved):

stuqdog (approved):

micheal-parks (approved):

njooma (approved):

PR #642, opened 2025-09-30
CONSULT-1797-implement-FileUpload-on-the-typescript-SDK

nicksanford:


stuqdog (approved):
A couple small things and we should make sure tests are passing (no need to rerequest review), otherwise lgtm!

src/app/data-client.ts
```diff
@@ -57,6 +64,43 @@ interface TabularDataPoint {
...
+   * Optional type of the component associated with the file (for example,
+   * "movement_sensor").
+   */
+  componentType?: string;
+
+  /** Optional name of the component associated with the file. */
+  componentName?: string;
+
+  /** Optional name of the method associated with the file. */
+  methodName?: string;
```
nicksanford, line 79: remove, these are invalid on arbitrary data uploads

src/app/data-client.ts
```diff
@@ -57,6 +64,43 @@ interface TabularDataPoint {
...
+
+  /**
+   * Optional name of the file. The empty string `""` will be assigned as the
+   * file name if one isn't provided.
+   */
+  fileName?: string;
```
nicksanford, line 85: this is required

stuqdog (approved):
one minor comment, otherwise lgtm (modulo Nick's comments)

src/app/data-client.spec.ts
```diff
@@ -1716,3 +1723,93 @@ describe('DataPipelineClient tests', () => {
...
   });
 });
+
+describe('fileUpload tests', () => {
+  const partId = 'testPartId';
+  const binaryData = new Uint8Array([1, 2, 3, 4, 5]);
+  const options: FileUploadOptions = {
+    componentType: 'componentType',
+    componentName: 'componentName',
+    methodName: 'methodName',
```
nicksanford, line 1733: these shouldn't be here

src/app/data-client.spec.ts
```diff
@@ -1716,3 +1723,93 @@ describe('DataPipelineClient tests', () => {
     });
   });
 });
+
+describe('fileUpload tests', () => {
+  const partId = 'testPartId';
+  const binaryData = new Uint8Array([1, 2, 3, 4, 5]);
```
nicksanford, line 1729: can you please test with data that is large enough to actually validate that your chunking logic works correctly?

src/app/data-client.ts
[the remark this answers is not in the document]
njooma: Ohhh if this is the case, lmk if there are other fields that are also unused/regretted and I can remove those from this SDK as well (and slate them for the other SDKs but that would be breaking)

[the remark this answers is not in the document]
dmhilly: Agree that this part of the API was done poorly and should be rationalized, however it is actually valid to set these fields for FileUpload requests and query on them. I see data that was uploaded with these options and am able to query it. So I think we should keep these options in for now, and at some point in the future the Data team will rationalize the many different upload APIs (it's on our eng plan) and how they should be used.

nicksanford (commented):
LGTM, ty @njooma for implementing this!

PR #654, opened 2025-10-23
only log max attempt when actually reached max attempts

DTCurrie:
Updates the post-disconnect reconnect backoff logic to log the 'max attempts reached' message only when we can determine the max attempts have actually been reached. Otherwise, we log the error.


benjirewis (approved):
Thanks! Seems good to give us more information for now. I'll be very curious what that error actually is.

PR #655, opened 2025-10-23
RSDK-12152 RSDK-12153 AudioIn and AudioOut component wrappers

oliviamiller:
adds AudioIn and AudioOut component types. Manually tested with a go server.


src/rpc/stream-client-stream.ts
```diff
@@ -62,7 +62,7 @@ export class StreamClientStream<
             success: resolve,
             failure: reject,
           };
-          this.startRequest();
+          this.startRequest(signal);
```
oliviamiller: needed to pass the abort signal to startRequest() here to be able to cancel the server-side stream

njooma (approved):

PR #665, opened 2025-11-06
breakup workflows

DTCurrie:
Just some minor workflow optimization, allowing us to cache npm, buf, and playwright artifacts/dependencies so they aren't regenerated (sometimes multiple times) per PR run.

Screenshot 2025-11-06 at 11 33 46 AM


DTCurrie:
will come back to this later

PR #672, opened 2025-11-07
Add observability when dropping/gathering ice candidates

DTCurrie:
No functionality changes, just a little bit of observability.

As we add more observability, we consider a way to enable/disable different levels of observability.

### DIFF
--- a/src/rpc/signaling-exchange.ts
+++ b/src/rpc/signaling-exchange.ts
@@ -259,11 +259,14 @@ export class SignalingExchange {
     if (this.exchangeDone || this.pc.iceConnectionState === 'connected') {
       if (event.candidate !== null) {
         // eslint-disable-next-line no-console
-        console.info('Dropping ICE candidate - exchange done or already connected', { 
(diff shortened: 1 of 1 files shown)


src/rpc/signaling-exchange.ts
```diff
@@ -255,23 +255,38 @@ export class SignalingExchange {
...
     if (this.callUuid === undefined || this.callUuid === '') {
+      // eslint-disable-next-line no-console
+      console.error(callUUIDUnset);
       throw new Error(callUUIDUnset);
     }
 
     if (event.candidate.candidate !== undefined) {
-      console.debug(`gathered local ICE ${event.candidate.candidate}`); // eslint-disable-line no-console
+      // eslint-disable-next-line no-console
+      console.info(`gathered local ICE ${event.candidate.candidate}`);
```
DTCurrie:
```suggestion
      console.info(`Gathered local ICE ${event.candidate.candidate}`);
```

PR #676, opened 2025-11-08
Handle exceptional errors when retrying connections

DTCurrie:
Previously, the `RobotClient`'s connection logic would attempt to retry on all errors, regardless of whether they were recoverable. This led to unnecessary retry cycles on fatal errors such as authentication failures, invalid arguments, or configuration issues.

This change introduces error classification into the client's connection and reconnection logic to prevent retries on non-recoverable errors.

### DIFF
--- a/src/robot/__fixtures__/dial-configs.ts
+++ b/src/robot/__fixtures__/dial-configs.ts
@@ -18,3 +18,13 @@ export const withNoReconnect: DialConf = {
   ...baseDialConfig,
   noReconnect: true,
 } as const;
+
+export const withRetry: DialConf = {
+  ...baseDialConfig,
+  noReconnect: false,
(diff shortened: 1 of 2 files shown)


src/robot/client.ts
```diff
@@ -130,6 +136,41 @@ const isPosInt = (x: number): boolean => {
...
+      Code.Canceled,
+      Code.InvalidArgument,
+      Code.NotFound,
+      Code.AlreadyExists,
+      Code.PermissionDenied,
+      Code.FailedPrecondition,
+      Code.OutOfRange,
+      Code.Unimplemented,
+      Code.Unauthenticated,
+    ].includes(error.code)
```
DTCurrie: Let me know if we should add/remove anything from this list.

PR #679, opened 2025-11-10
Fix session reset and add tests

DTCurrie:


DTCurrie:
closing, will reopen without all the conflicts

PR #690, opened 2025-11-24
APP-14421 TypeScript SDK support for CreateBinaryDataSignedURL

n0nick:


n0nick:
@njooma should I get #689 merged first before this code change?

PR #693, opened 2025-12-03
Automated Protos Update

viambot:
This is an auto-generated PR to update proto definitions. Check the commits to see which repos and commits are responsible for the changes


njooma:
@DTCurrie I didn't see that the tests had failed, that's my bad. Should be fixed

PR #694, opened 2025-12-11
RSDK-12724 finish e2e suite and general cleanuup

DTCurrie:
As we continue to refine our testing practices, we should ensure our testing setup aligns with them. This PR includes additional cleanup and restructuring to achieve that alignment and a better e2e suite that focuses on the current critical flows for the ts-sdk: robot connection, arms, cameras, and the vision service. 

If there are other critical flows we want to test, they should be easy to add, given this e2e test setup.

This is all just test file changes, so green should mean go. But check out how things are set up and structured. I will be adding a `STYLEGUIDE.md` file in the near future that covers testing best practices.


stuqdog (approved):
Didn't read super closely but the general changes all look reasonable to me!

PR #698, opened 2025-12-15
[RSDK-12928] - Add video service client

seanavery:
Simple PR to add client support for the new video service API

### DIFF
--- a/src/services/video/video.ts
+++ b/src/services/video/video.ts
@@ -19,7 +19,11 @@ export interface Video extends Resource {
    * );
    *
    * for await (const chunk of videoStream) {
-   *   console.log('Received video chunk:', chunk.videoData.length, 'bytes');
+   *   console.log(
+   *     'Received video chunk:',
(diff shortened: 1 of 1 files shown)


src/services/video/client.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    const stream = this.client.getVideo(request, callOptions);
+
+    // Yield each video chunk as it arrives from the server stream
+    for await (const response of stream) {
+      yield {
+        videoData: response.videoData,
+        videoContainer: response.videoContainer,
+        requestId: response.requestId,
+      };
```
njooma: I think you can yield the response directly, since it satisfies the `VideoChunk` interface

njooma (changes requested):

src/services/video/client.ts
seanavery: good call, changed

seanavery:
@njooma ready for re-review

PR #733, opened 2026-01-29
[RSDK-13285] Add global grpc trace logging

njooma:
Added trace logging with a global variable. To enable, you have to add `globalThis.VIAM = { GRPC_TRACE_LOGGING: true };` somewhere in your application.

Using the vanilla example in the TS SDK, the output looks like:

image

I decided on `trace` so that users might be able to follow the path of function calls

### DIFF
--- a/src/rpc/dial.ts
+++ b/src/rpc/dial.ts
@@ -134,7 +134,7 @@ const enableGRPCTraceLogging = <T extends Transport>(
         contextValues
       );
       // eslint-disable-next-line no-console
-      console.debug(`Unary response ${id} : ${JSON.stringify(resp)}`);
+      console.debug(`Unary response received: ${id}`);
       return resp;
     };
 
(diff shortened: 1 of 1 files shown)


njooma:
As implemented, this does NOT include responses in the trace log.

src/rpc/dial.ts
```diff
@@ -131,6 +133,9 @@ const enableGRPCTraceLogging = <T extends Transport>(
         message,
         contextValues
       );
+      // eslint-disable-next-line no-console
+      console.trace(`Unary response ${id} : ${JSON.stringify(resp)}`);
```
dgottlieb: I like having the stacks for the requests. In case there is some unexpected retry bug somewhere. IIUC, matching up the ID log lines results in stacks that are the same. If that's right, can turn this into a non-stack trace log line.

cheukt (approved):
lgtm

njooma:
merging for now to get it out in the imminent release

PR #747, opened 2026-02-12
Tweaks to Claude workflows

ale7714:
* Fix permissions, move permissions to the jobs that need it.


ale7714:
I'm gonna force merge because this doesn't impact the SDK and because I have to test in main for the end to end flow.

PR #781, opened 2026-02-20
Remove rdkenv/canon containers from CI workflows 

ale7714:
After looking into it, there's no reason for the TS sdk to use these images and it just seems to make workflows a bit more complex and confusing. The only explanation I have is either we copied the pattern from rdk or similar, and/or at some point we were using different runners that made this convenient.

Anyways i tested my change on the test workflow here https://github.com/viamrobotics/viam-typescript-sdk/actions/runs/22227605557/job/64298269141

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,4 +13,4 @@ jobs:
 
       - name: Build, lint, and test
         run: |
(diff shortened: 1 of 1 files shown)


ale7714:
@claude

PR #830, opened 2026-03-13
Fix check-examples script

allisonschiang:


scripts/check-examples.sh
```diff
@@ -5,15 +5,20 @@ failed=""
...
+		fi
+		if ! install_out=$(cd "$dir" && $install_cmd 2>&1); then
 			echo "$dir npm install failed:"
 			echo "$install_out" | grep -A2 "npm error" | head -20
 			failed="$failed $dir"
 			continue
 		fi
-		tsc_out=$(cd "$dir" && npx tsc --noEmit --pretty 2>&1)
-		if [ $? -ne 0 ]; then
+		if ! tsc_out=$(cd "$dir" && npx tsc --noEmit --pretty 2>&1); then
```
allisonschiang: if tsc fails, bash was killing the script so it was never getting to the output so we didn't know why it was failing.

PR #864, opened 2026-04-01
APP-15653: Add reconnecting and reconnection failed state

allisonschiang:
Proposal:
https://docs.google.com/document/d/1jwOwAHAoSIjSlPAYl-_H8yo2qBnoHr0uPItEnDLs4ck/edit?tab=t.0#heading=h.hloa3ivfpho6

Basically, instead of switching between connected and disconnected when trying to reconnect, it'll just emit one state during this entire time- reconnecting. Moreover, when this fails (retry attempts maxed out or it hits a non-retryable error) it will emit reconnection failed. 

This will not be merged until all parts are ready, but putting it up for feedback now

### DIFF
--- a/src/events.ts
+++ b/src/events.ts
@@ -10,6 +10,8 @@ export enum MachineConnectionEvent {
   DISCONNECTING = 'disconnecting',
   DISCONNECTED = 'disconnected',
   DIALING = 'dialing',
+  RECONNECTING = 'reconnecting',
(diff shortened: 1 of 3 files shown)


stuqdog (approved):
I haven't tested this myself but it looks reasonable to me!

allisonschiang:
Merging -> The app doesn't use the TS-sdk for connection directly (it uses the svelte sdk) so we can manage when to push that with the app changes

PR #898, opened 2026-04-14
Add CallOptions support to transformPose

danielbotros:
## Summary

- Adds an optional `callOptions?: CallOptions` parameter to `transformPose` in both the `Robot` interface and `RobotClient` implementation
- The options are passed through to the Connect RPC call, allowing callers to pass an `AbortSignal` (or other call options) to cancel in-flight requests at the HTTP/2 transport level

## Motivation

The motion tab in the web app polls `frameSystemConfig` every 5s and re-calls `transformPose` for every frame on each poll. Without abort support, slow network conditions can cause batches to accumulate — each poll's N in-flight requests pile up on top of the previous batch's, eventually hitting RDK's per-client inflight request cap of 100 and triggering rate limit errors.

## Follow-up

- viamrobotics/app#11709 — uses this to wire up `AbortController` in the motion tab effect, cancelling the previous batch of `transformPose` calls before each new poll starts

## Test plan

- [x] Verify `transformPose` still works normally without passing call options
- [x] Verify passing `{ signal: abortedSignal }` causes the call to reject immediately
- [x] Verify passing a signal and aborting mid-flight cancels the request

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/robot/__tests__/transform-pose.spec.ts
+++ b/src/robot/__tests__/transform-pose.spec.ts
@@ -1,0 +1,125 @@
+// @vitest-environment happy-dom
+
+import { describe, expect, it, vi } from 'vitest';
+import { RobotClient } from '../client';
+import * as rpcModule from '../../rpc';
+import { createRouterTransport } from '@connectrpc/connect';
+import { RobotService } from '../../gen/robot/v1/robot_connect';
+import { TransformPoseResponse } from '../../gen/robot/v1/robot_pb';
+import { PoseInFrame } from '../../types';
+import {
+  createMockDataChannel,
+  createMockPeerConnection,
+} from '../../__tests__/mocks/webrtc';
+import { withNoReconnect } from './fixtures/dial-configs';
+
+vi.mock('../../rpc', async () => {
+  const actual = await vi.importActual('../../rpc');
+  return {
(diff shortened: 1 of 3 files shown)


src/robot/client.ts
```diff
@@ -1259,14 +1260,15 @@ export class RobotClient extends EventDispatcher implements Robot {
...
+    supplementalTransforms: Transform[],
+    callOptions?: CallOptions
   ) {
     const request = new TransformPoseRequest({
       source,
       destination,
       supplementalTransforms,
     });
-    const response = await this.robotService.transformPose(request);
+    const response = await this.robotService.transformPose(request, callOptions);
```
danielbotros: Might have missed it elsewhere but couldn't find another example for what the correct pattern here is for robot service requests; passing in `callOptions` or the `AbortSignal` directly?

src/robot/__tests__/transform-pose.spec.ts
```diff
@@ -0,0 +1,125 @@
+// @vitest-environment happy-dom
```
danielbotros, line 1: Is there a better place for these tests?

njooma (approved):

PR #949, opened 2026-05-14
RSDK-10883 - add structured debug logging for connection and gRPC events

stuqdog:
Adds an opt-in debug logging system via `setDebugLogWriter`. When enabled, the SDK emits structured JSON entries for dial attempts (started/success/failed), gRPC requests and responses (including per-message stream logging), client disconnects, and ICE connectivity loss. Each connection is assigned a UUID so events can be correlated across a session.

### DIFF
--- a/src/debug.ts
+++ b/src/debug.ts
@@ -80,7 +80,9 @@ export const writeDebugLog = (
   event: string,
   fields: Record<string, unknown> = {}
 ): void => {
-  if (!currentWriter) return;
+  if (!currentWriter) {
+    return;
+  }
   currentWriter({
     timestamp: new Date().toISOString(),
     event,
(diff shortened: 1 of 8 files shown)


src/debug.ts
```diff
@@ -0,0 +1,113 @@
...
+ * - `grpc_request` — a gRPC request was sent through a connection
+ * - `grpc_response` — a gRPC response was received (includes `error` on failure)
+ * - `client_closed` — the client was explicitly closed
+ * - `ice_disconnected` — the WebRTC ICE connection entered the Disconnected state
+ */
+
+/** A single structured debug log entry emitted by the Viam SDK. */
+export interface DebugLogEntry {
+  /** ISO 8601 timestamp of when the event occurred. */
+  timestamp: string;
```
njooma: Any reason not to use native `Date` object here?

njooma (approved):

PR #981, opened 2026-07-10
Parse URDF kinematics files into structured joints/links

HipsterBrown:
## Summary

Builds on #978 (which stopped `getKinematics` from throwing on URDF/XML kinematics files) by actually **parsing** the URDF into the SDK's shared `KinematicsData` shape, so downstream UIs (joint-limit sliders, 3D viewers) get real structured data instead of just raw XML.

Brings in the `parseUrdf` converter authored in #977 and wires it into `getKinematicsFromClient`:

- **Valid URDF** -> parsed into SVA-unit `joints`/`links` (meters->mm, radians->degrees, fixed joints filtered, `continuous`->unbounded `revolute`, geometry mapped). Reported as `kinematic_param_type: 'SVA'` so consumers gated on `'SVA'` pick up joint limits, with the raw XML retained on the `urdf` field for origin/mesh resolution.
- **Malformed/empty URDF** -> degrades to #978's never-throw passthrough (`'URDF'`, empty joints/links, raw `urdf`). The never-throw guarantee is preserved.
- **SVA JSON / UNSPECIFIED** -> unchanged from #978.

`parseUrdf` and the kinematics types (`KinematicsData`, `GetKinematicsResult`, `GetKinematicsResultWithMeshes`) are now exported from the public API.

### Relationship to #977
This supersedes the code in #977 by rebasing its `parseUrdf` onto the merged #978 fix. `src/urdf.ts` is taken verbatim from #977 **except** one type fix: `orientationFromRpy` annotated its return `case` as `string`, widening the `'eulerAngles'` literal so it wasn't assignable to `Frame.orientation`. Narrowed to the literal (type-only; runtime unchanged). **This fix should also land in #977.**

### Notes / follow-ups
- Adds `fast-xml-parser` as a runtime dependency (~29 KB gzip). Because `parseUrdf` is a static public export, this lands in the bundle for all consumers; code-splitting it is an API-shape change best handled as a follow-up.
- Conversion paths for continuous/prismatic joints and the `meshesByUrdfFilepath` passthrough are not yet unit-tested (all `parseUrdf` internals from #977) -- candidate follow-up.

## Test Plan

- [x] `npm test` -- 409 passing (29 files), including new `getKinematicsFromClient` cases: valid URDF -> SVA joints/links with converted degree limits + retained `urdf`; malformed URDF -> never-throw fallback.
- [x] `src/urdf.spec.ts` -- parseUrdf conversion suite (name, fixed-joint exclusion, rad->deg, link poses, mesh/box geometry, throws on missing `<robot>`).
- [x] `tsc` -- 0 errors; `npm run build` clean; `eslint` clean.
- [ ] Verify against a real arm configured with a URDF kinematics file that joint-limit sliders populate downstream.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01EBrD9CPVH6CgGyQwo9cMjW

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -440,8 +440,6 @@ export {
   type KinematicsData,
 } from './utils';
 
-export { parseUrdf } from './urdf';
-
 export {
   createConsoleLogWriter,
   setDebugLogWriter,
--- a/src/main.ts
+++ b/src/main.ts
@@ -435,7 +435,12 @@ export {
   doCommandFromClient,
   enableDebugLogging,
   getStatusFromClient,
+  type GetKinematicsResult,
+  type GetKinematicsResultWithMeshes,
+  type KinematicsData,
 } from './utils';
+
+export { parseUrdf } from './urdf';
 
 export {
   createConsoleLogWriter,
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,7 +13,8 @@
         "@connectrpc/connect": "^1.7.0",
         "@connectrpc/connect-web": "^1.7.0",
         "bsonfy": "^1.0.2",
-        "exponential-backoff": "^3.1.3"
+        "exponential-backoff": "^3.1.3",
+        "fast-xml-parser": "^5.8.0"
       },
       "devDependencies": {
         "@bufbuild/buf": "^1.70.0",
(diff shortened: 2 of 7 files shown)


src/main.ts
```diff
@@ -435,8 +435,13 @@ export {
   doCommandFromClient,
   enableDebugLogging,
   getStatusFromClient,
+  type GetKinematicsResult,
+  type GetKinematicsResultWithMeshes,
+  type KinematicsData,
 } from './utils';
 
+export { parseUrdf } from './urdf';
```
njooma: Do we need to export it publicly? Seems that this might be an internal-only method? Do you foresee parsing urdf externally?

njooma (approved):
Just one question but looks good

src/main.ts
HipsterBrown: If we ever support authoring modules using Node.js, I can see it being useful. I'll keep it internal until we're ready to support it for external use. Thanks for reviewing!

